PR #2, opened 2023-11-22
feat/uploader

adriencaccia:


Cargo.toml
```diff
@@ -16,10 +16,20 @@ lazy_static = "1.4.0"
 log = "0.4.20"
 rand = "0.8.5"
 regex = "1.10.2"
-reqwest = { version = "0.11.22", features = ["json", "blocking"] }
+reqwest = { version = "0.11.22", features = ["json", "stream", "blocking"] }
```
art049: do we still need the blocking feature since we now use tokio?

Cargo.toml
```diff
@@ -16,10 +16,20 @@ lazy_static = "1.4.0"
...
+reqwest-middleware = "0.2.4"
+reqwest-retry = "0.3.0"
 serde = { version = "1.0.192", features = ["derive"] }
 serde_json = "1.0.108"
 url = "2.4.1"
+sha256 = "1.4.0"
+tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
```
art049: I think just `rt` is enough since we're not doing some heavy multi threading stuff (afaik)
Some interesting docs: https://rcos.io/static/internal_docs/tokio/index.html#feature-flags

Cargo.toml
```diff
@@ -16,10 +16,20 @@ lazy_static = "1.4.0"
...
+tokio-tar = "0.3.1"
+md5 = "0.7.0"
+temp-env = { version = "0.3.6", features = ["async_closure"] }
+insta = { version = "1.29.0", features = ["json", "redactions"] }
```
art049: Those 2 are dev-dependencies so we should move them in dev deps

src/app.rs
```diff
@@ -13,7 +13,7 @@ fn show_banner() {
 "#,
         VERSION
     );
-    println!("{}", banner);
+    info!("{}", banner);
```
art049: I didn't mention it when we spoke but the banner shouldn't be in the logs, that's why I just added a println

art049: otherwise formating will be uglier for this

src/uploader/ci_provider.rs
```diff
@@ -0,0 +1,13 @@
...
+use super::interfaces::UploadMetadata;
+
+pub trait CIProvider {
+    fn get_provider_name(&self) -> &'static str;
+
+    fn get_provider_slug(&self) -> &'static str;
+
+    fn detect() -> bool;
+
+    fn get_upload_metadata(&self, config: &Config, archive_hash: &str) -> Result<UploadMetadata>;
```
art049: Adding some docstring with small examples on the trait methods would be helpful

src/uploader/github_actions_provider.rs
```diff
@@ -0,0 +1,261 @@
...
+            event: env::var("GITHUB_EVENT_NAME").expect("GITHUB_EVENT_NAME"),
+            gh_data: GhData {
+                job: env::var("GITHUB_JOB").expect("GITHUB_JOB"),
+                run_id: env::var("GITHUB_RUN_ID").ok().unwrap().parse().unwrap(),
+                sender: Some(Sender {
+                    login: env::var("GITHUB_ACTOR").expect("GITHUB_ACTOR"),
+                    id: env::var("GITHUB_ACTOR_ID").ok().unwrap().parse().unwrap(),
+                }),
+            },
+            base_ref: env::var("GITHUB_BASE_REF").ok(),
```
art049: instead of unwraps and expects, we should return an err if something goes wrong here

art049: and it applies also for the step before of this method

src/uploader/upload.rs
```diff
@@ -1,7 +1,169 @@
...
+    interfaces::{UploadData, UploadMetadata},
+};
+
+fn get_provider(config: &Config) -> Result<impl CIProvider> {
+    if GitHubActionsProvider::detect() {
+        let provider = GitHubActionsProvider::try_from(config)?;
+        return Ok(provider);
+    }
+
+    Err(anyhow!("No CI provider detected"))
```
art049: we could have use bail from anyhow as well but this is fine

src/uploader/upload.rs
```diff
@@ -1,7 +1,169 @@
...
+const UPLOAD_RETRY_COUNT: u32 = 3;
+
+async fn retrieve_upload_data(
+    config: &Config,
+    upload_metadata: &UploadMetadata,
+) -> Result<UploadData> {
+    let retry_policy = ExponentialBackoff::builder().build_with_max_retries(UPLOAD_RETRY_COUNT);
+    let client = ClientBuilder::new(Client::new())
+        .with(RetryTransientMiddleware::new_with_policy(retry_policy))
+        .build();
```
art049: Making the client with retry reusable would be nice also to add the ability to use it as well for the file upload.
It lacks the customuseragent as well

src/uploader/upload.rs
```diff
@@ -1,7 +1,169 @@
...
+pub async fn upload(config: &Config, run_data: &RunData) -> Result<()> {
+    let (archive_buffer, archive_hash) = get_profile_archive_buffer(run_data).await?;
+
+    let provider = get_provider(config)?;
+    debug!("CI provider detected: {:#?}", provider.get_provider_name());
+
+    let upload_metadata = provider.get_upload_metadata(config, &archive_hash)?;
+    debug!("Upload metadata: {:#?}", upload_metadata);
+    if upload_metadata.tokenless {
+        let hash = digest(json!(upload_metadata).to_string());
```
art049: using the full qualified name `sha256::digest` would make clearer the kind of hash we build here

Cargo.toml
adriencaccia: Seen together, the runner still used a blocking call. I changed it to async to remove `blocking`

PR #5, opened 2023-11-28
feat/buildkite provider

adriencaccia:
- feat(provider): change implem of get_provider to allow different providers
- fix(github_actions): fix fork implementation
- feat(provider): use enum for run event instead of strings
- feat(provider): implement builkite provider


src/ci_provider/buildkite/provider.rs
```diff
@@ -0,0 +1,336 @@
...
+    static ref GITHUB_URL_REGEX: Regex = Regex::new(
+        r"(?x)
+            (?:https://github.com/|git@github.com:)
+            (?P<owner>[^/]+)/(?P<repository>[^/.]+)\.git
+        "
+    )
+    .expect("Failed to compile GitHub URL regex");
+}
+
+impl BuildkiteProvider {
```
art049: Is it necessary to bind those functions to the BuildkiteProvider Struct? They could live outside since all of then (almost) are static functions

src/ci_provider/buildkite/provider.rs
```diff
@@ -0,0 +1,336 @@
...
+                get_env_variable("BUILDKITE_ORGANIZATION_SLUG")?,
+                get_env_variable("BUILDKITE_PIPELINE_SLUG")?,
+            ),
+        })
+    }
+}
+
+impl CIProviderDetector for BuildkiteProvider {
+    fn detect() -> bool {
+        // check if the BUILDKITE environment variable is set and the value is truthy
```
art049:
```suggestion
```

art049: No comment needed imo and it's not exactly checking if it's truthy (`1` would not pass)

src/ci_provider/provider.rs
```diff
@@ -2,6 +2,17 @@ use crate::config::Config;
...
 use crate::uploader::UploadMetadata;
 
+pub trait CIProviderDetector {
+    /// Detects if the current environment is running inside the CI provider.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// assert_eq!(MyCIProvider::detect(), true);
+    /// ```
```
art049: not sure this example helps a lot here

src/runner/setup.rs
```diff
@@ -7,23 +7,33 @@ use crate::prelude::*;
 
 const VALGRIND_CODSPEED_VERSION: &str = "3.21.0-0codspeed1";
 
+/// Run a command with sudo if available
+fn run_with_sudo(command_args: Vec<&str>) -> Result<()> {
+    let use_sudo = Command::new("sudo").output().is_ok();
```
art049: hmm, I'm not sure this is the best method of checking if sudo is possible

src/runner/setup.rs
```diff
@@ -7,23 +7,33 @@ use crate::prelude::*;
 
 const VALGRIND_CODSPEED_VERSION: &str = "3.21.0-0codspeed1";
 
+/// Run a command with sudo if available
+fn run_with_sudo(command_args: Vec<&str>) -> Result<()> {
```
art049: Probably, this should return the built command without running it, giving the flexibility to anyone using the function to chose how the execution should go (ex: maybe I want to throw away stdout?)

PR #8, opened 2023-12-04
fix/codspeed rust

adriencaccia:
Control the environment when running `cargo-codspeed`, will fix performance regressions in https://github.com/astral-sh/ruff/pull/8989

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -16,6 +16,21 @@ pub struct Config {
     pub skip_setup: bool,
 }
 
+#[cfg(test)]
+impl Config {
+    /// Constructs a new `Config` with default values for testing purposes
+    pub fn test() -> Self {
+        Self {
+            upload_url: Url::parse(DEFAULT_UPLOAD_URL).unwrap(),
+            token: None,
+            working_directory: None,
+            command: "".into(),
+            skip_upload: false,
+            skip_setup: false,
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -16,6 +16,21 @@ pub struct Config {
     pub skip_setup: bool,
 }
 
+// Provide a default implementation for tests only
+#[cfg(test)]
+impl Default for Config {
+    fn default() -> Self {
```
art049: Implementing default only for tests is confusing.
Better create another factory, like Config::test()

src/runner/valgrind.rs
```diff
@@ -46,6 +46,17 @@ lazy_static! {
     };
 }
 
+fn get_bench_command(config: &Config) -> String {
+    let mut bench_command = config.command.clone();
+    bench_command = bench_command
+        // Fixes a compatibility issue with cargo 1.66+ running directly under valgrind <3.20
+        .replace("cargo codspeed", "cargo-codspeed")
+        // Escape double quotes since we're going to run this command in a subshell
+        .replace('"', "\\\"");
```
art049: We need to make sure this doesn't break anything, a bit scary

adriencaccia: I wanted to try, but it did not work; let's remove it

PR #18, opened 2024-04-22
feat: add actual_commit_hash ProviderMetadata property

adriencaccia:


adriencaccia:
@art049 you can review the aggregated file changes directly, I will squash in a single commit at the end

art049 (approved):
As discussed, we should replace `commit_hash` with the commit hash from libgit and remove this `actual_commit_hash`.
The rest of the logic to find the "actual" commit will be handled on our backend.

art049 (approved):

PR #20, opened 2024-04-29
fix: retrieve root_repository_path from git dir

adriencaccia:
Ensure that `root_repository_path` is inferred from the current working directory first, instead of hardcoding it depending on the provider.
This will ensure that the action picks up the git repository, not matter where the repository is actually cloned on the worker.

### DIFF
--- a/src/helpers/find_repository_root.rs
+++ b/src/helpers/find_repository_root.rs
@@ -1,0 +1,66 @@
+use std::path::{Path, PathBuf};
+
+// during normal execution, we want to find the repository root by looking for a .git directory
+// during tests, we want `find_repository_root` to always return `None`, so that we don't have to
+// create a git repository for each test
+
+#[cfg(not(test))]
+pub fn find_repository_root(base_dir: &Path) -> Option<PathBuf> {
+    _find_repository_root(base_dir)
+}
+
+#[cfg(test)]
+pub fn find_repository_root(_base_dir: &Path) -> Option<PathBuf> {
+    None
+}
+
+// the core logic is extracted into a separate function so that it can be tested
+fn _find_repository_root(base_dir: &Path) -> Option<PathBuf> {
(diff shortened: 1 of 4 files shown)


src/helpers/find_repository_root.rs
adriencaccia, line 1: @art049 I want this function to return `None` during tests so that the path does not depend on where the tests are run.
I did it that way, but here I am missing the opportunity to actually test the function. Do you have an idea on how to test it, while still stubbing it out during other tests?

adriencaccia: Seen together, move the logic to a new `_find_repository_root` function and test it. `find_repository_root` will just call it.
Also, pass the current directory as a parameter to the function and return a `PathBuf` instead of a `String`

src/helpers/find_repository_root.rs
```diff
@@ -0,0 +1,61 @@
...
+    let current_dir = base_dir.canonicalize().ok()?;
+
+    for ancestor in current_dir.ancestors() {
+        let git_dir = ancestor.join(".git");
+        if git_dir.exists() {
+            return Some(ancestor.to_path_buf());
+        }
+    }
+
+    log::debug!("Could not find repository root");
```
art049: This should be a warning/error level log

src/helpers/find_repository_root.rs
```diff
@@ -0,0 +1,61 @@
...
+pub fn find_repository_root(base_dir: &Path) -> Option<PathBuf> {
+    _find_repository_root(base_dir)
+}
+
+#[cfg(test)]
+pub fn find_repository_root(_base_dir: &Path) -> Option<PathBuf> {
+    None
+}
+
+fn _find_repository_root(base_dir: &Path) -> Option<PathBuf> {
```
art049: We need a small comment that explain this whole thing with cfg test and so on

art049 (approved):

PR #39, opened 2024-09-17
cod 177 run time based benchmarks in the codspeed runner

adriencaccia:
- **refactor(runner): create executor instrument abstraction and add ValgrindInstrument**
- **feat(runner): add WallTimeInstrument**
- **feat(runner): add cpu and memory data to SytemInfo**


src/run/runner/helpers/run_command_with_log_pipe.rs
```diff
@@ -0,0 +1,52 @@
...
+                    String::from_utf8_lossy(&buffer[..bytes_read])
+                );
+            });
+        }
+        Ok(())
+    }
+
+    let mut process = cmd
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
```
fargito, line 37: Do we redirect stderr to stdin because of a Github actions runner specificity?

src/run/runner/valgrind/measure.rs
```diff
@@ -0,0 +1,99 @@
...
+) -> Result<()> {
+    // Create the command
+    let mut cmd = Command::new("setarch");
+    cmd.arg(ARCH).arg("-R");
+    // Configure the environment
+    cmd.envs(BASE_INJECTED_ENV.iter()).env(
+        "PATH",
+        format!(
+            "{}:{}",
+            setup_introspected_node()
```
fargito:
```suggestion
            setup_introspected_nodejs()
```
Maybe a little bit more explicit

src/run/check_system.rs
```diff
@@ -50,12 +60,37 @@ impl SystemInfo {
...
             os,
             os_version,
             arch,
             host,
             user,
+            cpu_brand,
+            cpu_name,
+            cpu_vendor_id,
+            cpu_cores,
+            total_memory_gb: ram_quantity,
```
art049: this could be named `total_memory_gb` directly

src/run/runner/valgrind/instrument.rs
```diff
@@ -0,0 +1,52 @@
...
+            None
+        };
+        measure::measure(config, profile_folder, &mongo_tracer)?;
+        harvest_perf_maps(profile_folder)?;
+        if let Some(mut mongo_tracer) = mongo_tracer {
+            mongo_tracer.stop().await?;
+        }
+        end_group!();
+
+        Ok(())
```
art049: A lot of this logic should not be directly here and should probably be relocated to `src/run/mod.rs` if we want to make an abstract executor(and then reuse most of the logic we already have for another instrument)

- the groups for should be managed outside of the executor
- we should create a dedicated setup method in the executor
- the executor itself should not start/stop the other instruments

src/run/runner/mod.rs
```diff
@@ -1,8 +1,22 @@
...
+use anyhow::bail;
+use executor_instrument::ExecutorInstrument;
+use executor_instrument::ExecutorInstrumentDetector;
+use valgrind::instrument::ValgrindInstrument;
+
+pub use executor_instrument::RunData;
 pub use valgrind::VALGRIND_EXECUTION_TARGET;
+
+pub fn get_executor_instrument() -> Result<Box<dyn ExecutorInstrument>> {
+    if ValgrindInstrument::detect() {
```
art049: I think this `ExecutorInstrumentDetector` trait is overkill. For now, we have at most two executor instruments having the logic here or in `src/run/mod.rs` would be simpler

src/run/runner/executor_instrument.rs
```diff
@@ -0,0 +1,33 @@
...
+        profile_folder: &Path,
+    ) -> Result<()>;
+
+    /// Runs the executor instrument
+    async fn run(&self, config: &Config, system_info: &SystemInfo) -> Result<RunData> {
+        let profile_folder = create_profile_folder()?;
+        self.run_executor(config, system_info, &profile_folder)
+            .await?;
+        Ok(RunData { profile_folder })
+    }
```
art049: In my opinion, this logic should be directly in `src/run/mod.rs`
The difference between run and run_executor is not super easy to understand. I think we shouldn't use the trait system to share code in our case, but we should prefer our parent function calling the executor

src/run/runner/mod.rs
```diff
@@ -7,25 +7,44 @@ mod wall_time;
...
 pub use valgrind::VALGRIND_EXECUTION_TARGET;
-use wall_time::instrument::WallTimeInstrument;
+pub use wall_time::instrument::WallTimeInstrument;
 
-pub fn get_executor_instrument() -> Result<Box<dyn ExecutorInstrument>> {
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq, Hash)]
+pub enum ExecutorInstrumentName {
+    Valgrind,
+    Macro,
+}
```
art049: IMO this could be a trait variable in each instrument, so the `get_active_executor_name` is not needed, and we decide on the instrument once

art049 (changes requested):
I think we need to refactor a bit the executor instrument and even if it might be a big change right now we can keep it for the future instruments

src/run/runner/helpers/run_command_with_log_pipe.rs
adriencaccia: We redirect both stdout and stderr to the logger, and it's up to the logger to decide what to do with the logs. We have a logger for each `ci_provider` (GitHub Actions, Buildkite, and the local provider).

src/run/runner/mod.rs
```diff
@@ -1,8 +1,29 @@
...
+pub use interfaces::{ExecutorName, RunData};
+use valgrind::executor::{ValgrindExecutor, VALGRIND_RUNNER_MODE};
+
 pub use valgrind::VALGRIND_EXECUTION_TARGET;
+
+pub fn get_executor() -> Result<Box<dyn Executor>> {
+    match env::var("CODSPEED_RUNNER_MODE").as_deref() {
+        // if the mode is valgrind, or the env var is not set, use valgrind
+        Ok(VALGRIND_RUNNER_MODE) | Err(_) => Ok(Box::new(ValgrindExecutor)),
+        Ok(_) => bail!("Unknown codspeed runner mode"),
```
art049: maybe more simple is possible here

art049 (approved):

PR #43, opened 2024-10-10
chore: support ubuntu 24.04

art049:


adriencaccia (commented):
You can update the readme as well https://github.com/CodSpeedHQ/runner/blob/99a6193da6553227e1fc2676c83ce2daa44f9890/README.md?plain=1#L38

art049:
Cancelled the merge because we have a beta right now so I prefer using `v3-maintenance` while the beta is not 100% safe

PR #52, opened 2024-11-27
fix(gitlab): use correct ref for tag pipelines

fargito:


adriencaccia (approved):
Waiting on @art049

art049 (approved):
LGTM, we need to remove those beta commits though since it's changing the changelog and creates useless noise.

Will need to address this issue separately

PR #57, opened 2025-01-15
feat: send `RunPart` metadata to upload endpoint

fargito:


fargito:
@adriencaccia ready for a first review

.vscode/settings.json
```diff
@@ -6,6 +6,8 @@
   "[dotenv][ignore][properties][shellscript]": {
     "editor.defaultFormatter": "foxundermoon.shell-format"
   },
+  "files.insertFinalNewline": true,
+  "files.trimFinalNewlines": true,
```
adriencaccia: Handling final newline is already taken care of by `pre-commit` in https://github.com/CodSpeedHQ/runner/blob/199967280fabf61ebed6aa2111536b30f0150040/.pre-commit-config.yaml#L9.

So let's not have this setting here, this is the kind of setting that's better if it lives in the users' personal settings

src/run/ci_provider/provider.rs
```diff
@@ -55,7 +57,10 @@ pub trait CIProvider {
     /// let provider = MyCIProvider::new();
     /// assert_eq!(provider.get_provider_slug(), "my-ci-provider");
     /// ```
-    fn get_provider_slug(&self) -> &'static str;
+    fn get_platform(&self) -> &'static str;
+
+    /// Returns the slug of the CI Provider (also called platform)
+    fn get_platform_slug(&self) -> PlatformSlug;
```
adriencaccia: Let's only keep `get_platform_slug` and send only this field

src/run/ci_provider/github_actions/provider.rs
```diff
@@ -137,6 +138,55 @@ impl CIProvider for GitHubActionsProvider {
...
+            let matrix_str = serde_json::to_string(&matrix).expect("Unable to re-serialize matrix");
+            let strategy_str =
+                serde_json::to_string(&strategy).expect("Unable to re-serialize strategy");
+
+            for (k, v) in matrix.into_iter() {
+                metadata.insert(k, v);
+            }
+            for (k, v) in strategy.into_iter() {
+                metadata.insert(k, v);
+            }
```
adriencaccia: We can simplify the code here:

```suggestion
            metadata.extend(matrix);
            metadata.extend(strategy);
```

src/run/ci_provider/github_actions/provider.rs
```diff
@@ -137,6 +138,55 @@ impl CIProvider for GitHubActionsProvider {
...
+        let gh_strategy = get_env_variable("GH_STRATEGY")
+            .ok()
+            .and_then(|v| serde_json::from_str::<Value>(&v).ok());
+
+        let run_part_id = if let (Some(Value::Object(matrix)), Some(Value::Object(strategy))) =
+            (gh_matrix, gh_strategy)
+        {
+            let matrix_str = serde_json::to_string(&matrix).expect("Unable to re-serialize matrix");
+            let strategy_str =
+                serde_json::to_string(&strategy).expect("Unable to re-serialize strategy");
```
adriencaccia: It is a bit weird to first parse the strings and then serialize them again.
Let's keep the raw strings in separate variables, and then try to serialize them into other variables.

adriencaccia (changes requested):

.github/workflows/ci.yml
```diff
@@ -23,3 +26,18 @@ jobs:
...
+      - name: Check for todos
+        id: check-todos
+        run: ./scripts/show-todos.sh
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@24bffb9b452ba05a4f3f77933840a6a841d1b32b
+        with:
+          message: |
+            ${{ steps.check-todos.outputs.pr-comment }}
+          comment-tag: execution
+          mode: ${{ steps.check-todos.outputs.pr-comment == '' && 'delete' || 'upsert' }}
```
art049: This should not be enabled on a public repo since these references are not accessible to external contributors

src/run/ci_provider/buildkite/provider.rs
```diff
@@ -127,10 +128,14 @@ impl CIProvider for BuildkiteProvider {
...
 
-    fn get_provider_slug(&self) -> &'static str {
+    fn get_platform(&self) -> &'static str {
         "buildkite"
     }
 
+    fn get_platform_slug(&self) -> PlatformSlug {
+        PlatformSlug::Buildkite
+    }
+
```
art049: This seems not correct. The meaning of slug here is a stringified unique identifier. 
So having a rust enum named PlatformSlug seem to be an antipattern. If serde does the slugification then, this should not be suffixed with slug.

art049: Also, what's the motivation behind renaming the provider into the platform ? And having both something to get the slug and something to get the enum(this seems redundant)?

src/run/ci_provider/provider.rs
art049: I agree that we should keep only one of those. Keeping the one returning the enum would be best as suggested by @adriencaccia but I think we should rename it to `get_platform` and the enum to just `Platform`

src/run/uploader/upload_metadata.rs
```diff
@@ -60,9 +61,9 @@ mod tests {
         };
 
         let hash = upload_metadata.get_hash();
-        assert_eq!(
+        assert_snapshot!(
             hash,
-            "161a1a3eeea6d988909142e1e7bae3339b3698aaeb025641aa63809895336ae7"
+            @"b672792aac09f4331b387bf66a5e9225b6649e1e3bedacc6fa20e83ceb9956d6"
         );
```
art049: I think this test has a counterpart in our backend to make sure our hashing is consistent on both side.
And that might be the reason behind this not being a snapshot already. If that's the case, we should keep this an assert_eq and add a comment mentioning the relation with our backend

src/run/ci_provider/github_actions/provider.rs
art049: Why do we need to reserialize them at all?

art049 (changes requested):
A few minor changes 💪 
Make sure to fixup without force-pushing so we can compare the diff easily

src/run/ci_provider/github_actions/provider.rs
fargito: The issue with the raw string is that it is multiline. As we need to use it compute an identifier, we need it as a single line. Plus we are interested in the content of these objects, so it makes sense to parse and re-serialize them. I agree that it is not explicit though, I'll add a comment in the code.

src/run/ci_provider/buildkite/provider.rs
fargito: As seen together, we will keep `platform` in the runner, but serialize it to `platformSlug` in order to match the changes in our DB.

src/run/uploader/upload_metadata.rs
fargito: I'll mention this here, and change the backend value in the related backend PR

fargito:
@adriencaccia @arthur I updated and also added some tests

fargito:
@art049 sorry about the force-pushes. Here is the comparison link between the reviewed and the latest version: https://github.com/CodSpeedHQ/runner/compare/f87e92ed46b24c5635282521365023e1a31571d5..ed81950a62afe82f0c8d2839fa940d46ea9bc5a3

adriencaccia (approved):

art049 (commented):
As discussed together, it would be best to just keep job-index and job-total in the last part of the matrix part id

art049 (approved):

fargito:
@adriencaccia I removed some values from the metadata, can you release a new beta version?

src/run/run_environment/provider.rs
```diff
@@ -43,22 +44,15 @@ pub trait CIProvider {
     ///
     /// ```
     /// let provider = MyCIProvider::new();
-    /// assert_eq!(provider.get_provider_name(), "MyCIProvider");
+    /// assert_eq!(provider.get_run_environment_name(), "MyCIProvider");
     /// ```
-    fn get_provider_name(&self) -> &'static str;
+    fn get_run_environment_name(&self) -> &'static str;
```
adriencaccia: Since this is solely used for debug logging, remove this in favor of directly using the `get_run_environment`

src/run/uploader/upload.rs
```diff
@@ -109,13 +113,16 @@ pub struct UploadResult {
...
-    provider: &Box<dyn CIProvider>,
+    provider: &Box<dyn RunEnvironmentProvider>,
     run_data: &RunData,
     executor_name: ExecutorName,
 ) -> Result<UploadResult> {
     let (archive_buffer, archive_hash) = get_profile_archive_buffer(run_data).await?;
 
-    debug!("CI provider detected: {:#?}", provider.get_provider_name());
+    debug!(
+        "Run Environment provider detected: {:#?}",
```
adriencaccia: Simplify this format with simply `{:?}`
And ensure that the displayed value is indeed a string representation of the enum value

src/run/run_environment/provider.rs
```diff
@@ -54,6 +54,9 @@ pub trait RunEnvironmentProvider {
     /// Returns the metadata related to the RunEnvironment.
     fn get_run_environment_metadata(&self) -> Result<RunEnvironmentMetadata>;
 
+    /// Return the metadata necessary to identify the `RunPart`
+    fn get_platform_run_part(&self) -> Option<RunPart>;
```
adriencaccia: Rename to `get_run_environment_run_part`

adriencaccia (changes requested):

src/run/uploader/upload.rs
fargito:
```
[DEBUG::codspeed::run::uploader::upload] Run Environment provider detected: Local
```

fargito: Looks ok

fargito:
@adriencaccia updated

adriencaccia (approved):
LGTM
Waiting on @art049

art049 (approved):

PR #59, opened 2025-02-05
fix: unify environments between the two modes

art049:
@adriencaccia there was confusion between Mode and Executor, so I refactored this logic. I hadn't pulled it, so I redid the changes you made in the last commit, but in another way.

### DIFF
--- a/src/run/runner/executor.rs
+++ b/src/run/runner/executor.rs
@@ -1,11 +1,8 @@
-use super::helpers::env::BASE_INJECTED_ENV;
 use super::interfaces::{ExecutorName, RunData};
 use crate::prelude::*;
 use crate::run::instruments::mongo_tracer::MongoTracer;
 use crate::run::{check_system::SystemInfo, config::Config};
 use async_trait::async_trait;
-use std::collections::HashMap;
-use std::path::Path;
 
 #[async_trait(?Send)]
 pub trait Executor {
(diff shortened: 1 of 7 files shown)


src/run/runner/valgrind/measure.rs
```diff
@@ -42,17 +43,20 @@ lazy_static! {
...
     mongo_tracer: &Option<MongoTracer>,
 ) -> Result<()> {
     // Create the command
     let mut cmd = Command::new("setarch");
-    cmd.envs(base_env);
     cmd.arg(ARCH).arg("-R");
     // Configure the environment
-    cmd.env(
+    cmd.envs(get_base_injected_env(
+        RunnerMode::Instrumentation,
```
adriencaccia, line 55: Should we have the executor define its `RunnerMode`?

So that we do not have to hardcode it here, and instead have it as a method of the `Executor` trait

src/run/runner/executor.rs
```diff
@@ -1,11 +1,8 @@
-use super::helpers::env::BASE_INJECTED_ENV;
 use super::interfaces::{ExecutorName, RunData};
 use crate::prelude::*;
 use crate::run::instruments::mongo_tracer::MongoTracer;
 use crate::run::{check_system::SystemInfo, config::Config};
```
GuillaumeLagrange, line 4: nit: import grouping

GuillaumeLagrange (approved):
lgtm, tested locally and it works, no more `invalid measurement mode: valgrind` 😎

src/run/runner/valgrind/measure.rs
art049: this is not useful with the valgrind executor since it needs refactoring to be properly implemented. We could drill the mode into `measure` but that would be pointless.

src/run/runner/executor.rs
art049: can cargo fmt handle this?

GuillaumeLagrange: we could set this in our repos: https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=granula#Crate%5C%3A

GuillaumeLagrange: nevermind it's unstable and therefore not able to be put in rustfmt.toml, it needs to be set at IDE level :cry:

art049: :(

PR #65, opened 2025-02-19
fix: fix valgrind version checks

art049:
Avoid using the deb package version since --version only outputs our version


GuillaumeLagrange (approved):
lgtm to me

PR #67, opened 2025-03-06
feat(runner): distinguish instrumented command and tool error in codspeed

not-matthias:
Added a small wrapper script, which executes `cargo codspeed run` in a subprocess and then writes the status code to a file. 

Logs when the benchmark failed (in this case with a `panic`):
```
[DEBUG::codspeed::run::runner::valgrind::measure] Valgrind exit code = Some(0), Program exit code = 1
Error: failed to execute the benchmark process
```

Logs when valgrind error is detected (only an example, this doesn't cause valgrind to crash)
```
[DEBUG::codspeed::run::runner::valgrind::measure] valgrind.log: ==486607== Using source line as position.
--486607-- warning: L3 cache found, using its data for the LL simulation.
--486607-- warning: specified LL cache: line_size 64  assoc 12  total_size 18,874,368
--486607-- warning: simulated LL cache: line_size 64  assoc 18  total_size 18,874,368
==486607== 
==486607== Process terminating with default action of signal 4 (SIGILL): dumping core
==486607==  Illegal opcode at address 0x1205CF
==486607==    at 0x1205CF: my_bench::fibonacci (in /app/rust-crate/my_bench)
==486607==    by 0x120E57: codspeed_divan_compat::compat_divan::compat::bench::args::bench (in /app/rust-crate/my_bench)
==486607==    by 0x12313B: codspeed_divan_compat::compat_divan::compat::main (in /app/rust-crate/my_bench)
==486607==    by 0x121362: std::sys::backtrace::__rust_begin_short_backtrace (in /app/rust-crate/my_bench)
==486607==    by 0x121358: std::rt::lang_start::{{closure}} (in /app/rust-crate/my_bench)
==486607==    by 0x140DDF: std::rt::lang_start_internal (in /app/rust-crate/my_bench)
==486607==    by 0x12067B: main (in /app/rust-crate/my_bench)

Error: failed to execute valgrind
```

### DIFF
--- a/src/run/runner/valgrind/measure.rs
+++ b/src/run/runner/valgrind/measure.rs
@@ -9,8 +9,12 @@ use crate::run::{config::Config, instruments::mongo_tracer::MongoTracer};
 use lazy_static::lazy_static;
 use std::env;
 use std::fs::canonicalize;
+use std::io::Write;
+use std::os::unix::fs::PermissionsExt;
+use std::os::unix::process::ExitStatusExt;
 use std::path::Path;
 use std::{env::consts::ARCH, process::Command};
+use tempfile::TempPath;
 
 lazy_static! {
     static ref VALGRIND_BASE_ARGS: Vec<String> = {
@@ -40,6 +44,33 @@ lazy_static! {
         ));
         args
     };
+}
+
+/// Creates the shell script on disk and returns the path to it.
+fn create_run_script() -> anyhow::Result<TempPath> {
+    // The command is wrapped in a shell script, which executes it in a
+    // subprocess and then writes the exit code to a file. The process will
+    // always exit with status code 0, unless valgrind fails.
+    //
+    // Args:
+    // 1. The command to execute
+    // 2. The path to the file where the exit code will be written
+    const WRAPPER_SCRIPT: &str = r#"#!/bin/sh
+           (eval $1)
+           status=$?
+           echo -n "$status" > "$2"
+            "#;
+
+    let rwx = std::fs::Permissions::from_mode(0o777);
+    let mut script_file = tempfile::Builder::new()
+        .suffix(".sh")
+        .permissions(rwx)
+        .tempfile()?;
+    script_file.write_all(WRAPPER_SCRIPT.as_bytes())?;
+
+    // Note: We have to convert the file to a path to be able to execute it.
+    // Otherwise this will fail with 'File is busy' error.
+    Ok(script_file.into_temp_path())
 }
 
 pub fn measure(
@@ -85,8 +116,14 @@ pub fn measure(
         .arg(format!("--callgrind-out-file={}", profile_path.to_str().unwrap()).as_str())
         .arg(format!("--log-file={}", log_path.to_str().unwrap()).as_str());
 
-    // Set the command to execute
-    cmd.args(["sh", "-c", get_bench_command(config)?.as_str()]);
+    // Set the command to execute:
+    let script_path = create_run_script()?;
+    let cmd_status_path = tempfile::NamedTempFile::new().unwrap().into_temp_path();
+    cmd.args([
+        script_path.to_str().unwrap(),
+        get_bench_command(config)?.as_str(),
+        cmd_status_path.to_str().unwrap(),
+    ]);
 
     // TODO: refactor and move this to the `Instruments` struct
     if let Some(mongo_tracer) = mongo_tracer {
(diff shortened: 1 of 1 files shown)


src/run/runner/valgrind/measure.rs
```diff
@@ -96,7 +107,30 @@ pub fn measure(
     debug!("cmd: {:?}", cmd);
     let status = run_command_with_log_pipe(cmd)
         .map_err(|e| anyhow!("failed to execute the benchmark process. {}", e))?;
+    let cmd_status = {
+        let mut content = String::new();
+        cmd_status_file.read_to_string(&mut content)?;
+
+        content.parse::<u32>().unwrap_or(1)
```
art049: If we can't parse this, I think this should fail without overriding the error code

not-matthias:
Depends on https://github.com/CodSpeedHQ/codspeed-rust/pull/86

The correct exit code is now displayed (illegal instruction = signal 4)
```
[DEBUG::codspeed::run::runner::valgrind::measure] Valgrind exit code = Some(0), Valgrind signal = None, Program exit code = 132
Error: failed to execute the benchmark process, exit code: 132
```

art049:
Is there an easy way for us to write a test of this behavior?

src/run/runner/valgrind/measure.rs
```diff
@@ -42,6 +46,63 @@ lazy_static! {
...
+        script_file.write_all(WRAPPER_SCRIPT.as_bytes())?;
+
+        // Note: We have to convert the fiel to a path to be able to execute it.
+        // Otherwise this will fail with 'File is busy' error.
+        script_file.into_temp_path()
+    };
+
+    let cmd_status_path = tempfile::NamedTempFile::new().unwrap().into_temp_path();
+    let run_cmd = format!(
+        "{} \"{}\" \"{}\"",
```
art049: Defining the command as a string like this seems unsafe. We had many issues with the underlying command already containing quotes and escaped quotes. This quote escaping was handled by `Command` directly. Now, what will happen if we have quotes or escaped quotes in this?

src/run/runner/valgrind/measure.rs
```diff
@@ -84,9 +145,9 @@ pub fn measure(
         )
         .arg(format!("--callgrind-out-file={}", profile_path.to_str().unwrap()).as_str())
         .arg(format!("--log-file={}", log_path.to_str().unwrap()).as_str());
-
-    // Set the command to execute
-    cmd.args(["sh", "-c", get_bench_command(config)?.as_str()]);
+    // Set the command to execute:
+    let (run_cmd, _script, cmd_status_file) = safe_run_cmd(get_bench_command(config)?.as_str())?;
```
art049: If we don't need the script path, we shouldn't include it in the result.

src/run/runner/valgrind/measure.rs
```diff
@@ -42,6 +46,63 @@ lazy_static! {
...
+            echo -n "$status" > "$2"
+        }
+
+        # Args:
+        # 1. Cmd that should be executed
+        # 2. File to write the status code to
+        if [ "$#" -ne 2 ]; then
+            echo "Usage: $0 <cmd> <file>"
+            exit 1
+        fi
```
art049: I'm not sure we want to include usage guidelines in this since if we fail here, we will have much bigger problems. This could be some comments directly in the rust code

src/run/runner/valgrind/measure.rs
```diff
@@ -96,9 +157,62 @@ pub fn measure(
...
+mod tests {
+    use super::*;
+
+    fn safe_run(to_execute: &str) -> String {
+        let (run_cmd, _script, out_status) = safe_run_cmd(to_execute).unwrap();
+
+        let mut cmd = Command::new("sh");
+        cmd.args(["-c", &run_cmd]);
+
+        assert!(cmd.output().unwrap().status.success());
```
art049: Checking the status here and not in the test cases directly is a bit confusing

art049 (changes requested):
I'm not super confident merging this right now, mostly because of the quoting and wrapping most parameters. 

In
```
cmd.args(["sh", "-c", &run_cmd]);
```

I think we should replace `sh -c` with our wrapper script directly. I think it would avoid messing with the quotes and formatting command as strings.

src/run/runner/valgrind/measure.rs
not-matthias: Since it is written to a temp file, we have to keep the `TempPath` in scope, because it is deleted on `Drop`. According to the [docs](https://docs.rs/tempfile/latest/tempfile/struct.TempPath.html), we could set `.keep(true)` to not delete it.

src/run/runner/valgrind/measure.rs
```diff
@@ -42,6 +46,38 @@ lazy_static! {
...
+    // 2. The path to the file where the exit code will be written
+    const WRAPPER_SCRIPT: &str = r#"#!/bin/sh
+            safe_run()
+            {
+               (eval $1)
+               status=$?
+               echo -n "$status" > "$2"
+            }
+
+            safe_run "$1" "$2"
```
art049: I think you can even remove the function here so we never quote the command
```suggestion
           (eval $1)
           status=$?
           echo -n "$status" > "$2"
```

art049 (approved):
just one last thing to simplify the script and I think we're good to go

PR #72, opened 2025-03-14
fix(runner): check child status code after valgrind

not-matthias:
Prints the valgrind logs now: 
```
[DEBUG::codspeed::run::runner::valgrind::measure] Valgrind exit code = None, Valgrind signal = Some(11)
[DEBUG::codspeed::run::runner::valgrind::measure] valgrind.log: ==303374== Using source line as position.
--303374-- warning: L3 cache found, using its data for the LL simulation.
--303374-- warning: specified LL cache: line_size 64  assoc 12  total_size 18,874,368
--303374-- warning: simulated LL cache: line_size 64  assoc 18  total_size 18,874,368

Error: failed to execute valgrind
```

Before it only printed this: 
```
Error: cannot parse integer from empty string
```

### DIFF
--- a/src/run/runner/valgrind/measure.rs
+++ b/src/run/runner/valgrind/measure.rs
@@ -133,16 +133,10 @@ pub fn measure(
     debug!("cmd: {:?}", cmd);
     let status = run_command_with_log_pipe(cmd)
         .map_err(|e| anyhow!("failed to execute the benchmark process. {}", e))?;
-    let cmd_status = {
-        let content = std::fs::read_to_string(&cmd_status_path)?;
-        content.parse::<u32>()?
-    };
-
     debug!(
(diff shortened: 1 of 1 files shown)


src/run/runner/valgrind/measure.rs
```diff
@@ -155,6 +149,11 @@ pub fn measure(
     }
 
     // Check the exit code which was written to the file by the wrapper script.
+    let cmd_status = {
+        let content = std::fs::read_to_string(&cmd_status_path)?;
+        content.parse::<u32>()?
+    };
```
art049: maybe now this should just be a result and we should handle the error separately? like unable to retrieve program exit code?

not-matthias: Good idea, I'll fix that :+1:

art049 (approved):
Claud LGTM

PR #73, opened 2025-03-25
feat: add perf support in walltime

not-matthias:
TODOs: 
- [x] ~~Don't use max freq (or parse it and send it with the traces)~~ Actually doesn't matter. Higher freq, better callgraph. 
- [x] upload data + debug symbols to server
- [x] ~~use correct `codspeed` version (depends on https://github.com/CodSpeedHQ/codspeed-rust/pull/92)~~

Debug logs of successful run:
```
►►► Running the benchmarks
[DEBUG::codspeed::run::runner::helpers::setup] Running command: sudo sysctl -w kernel.kptr_restrict=0
[DEBUG::codspeed::run::runner::helpers::setup] Running command: sudo sysctl -w kernel.perf_event_paranoid=1
[DEBUG::codspeed::run::runner::wall_time::executor] cmd: ARCH="x86_64" CODSPEED_ENV="runner" CODSPEED_PROFILE_FOLDER="/tmp/profile.riowzUmLRF.out" CODSPEED_RUNNER_MODE="walltime" PYTHONHASHSEED="0" "sh" "-c" "perf record --data --freq=1000 --switch-output --control=fifo:/tmp/codspeed_perf.ctl.fifo,/tmp/codspeed_perf.ack.fifo --delay=-1 -g --output=/tmp/.tmpeI44Iq/perf.datao1Dpk6 -- cargo-codspeed run"
[DEBUG::codspeed::run::runner::wall_time::executor] Perf PID: 527912
Events disabled
[cargo-codspeed] Measurement mode: Walltime

Collected 1 benchmark suite(s) to run
Running graph-analysis my_bench
Timer precision: 13 ns
my_bench        fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ fibonacci                  │               │               │               │         │
Events enabled
[ perf record: dump data: Woken up 1 times ]
[ perf record: Dump /tmp/.tmpeI44Iq/perf.datao1Dpk6.2025032512120419 ]
Events disabled
│  ╰─ 32        4.563 ms      │ 7.109 ms      │ 4.881 ms      │ 5.07 ms       │ 100     │ 100
╰─ other_bench                │               │               │               │         │
Events enabled
[ perf record: dump data: Woken up 2 times ]
[ perf record: Dump /tmp/.tmpeI44Iq/perf.datao1Dpk6.2025032512120472 ]
Events disabled
   ╰─ 50        50.06 ms      │ 50.31 ms      │ 50.1 ms       │ 50.12 ms      │ 100     │ 100

Done running   ⠉ Running the benchmarks...                                                                                                                                   my_bench
Finished running 1 benchmark suite(s)
[ perf record: Woken up 2 times to write data ]
[ perf record: Dump /tmp/.tmpeI44Iq/perf.datao1Dpk6.2025032512120974 ]
[ perf record: Captured and wrote 0.034 MB /tmp/.tmpeI44Iq/perf.datao1Dpk6.<timestamp> ]
  ⠉ Running the benchmarks...                                                                                                                                                [src/run/runner/wall_time/executor.rs:117:9] &perf_files = [
    "/tmp/.tmpeI44Iq/perf.datao1Dpk6.2025032512120419",
    "/tmp/.tmpeI44Iq/perf.datao1Dpk6.2025032512120472",
    "/tmp/.tmpeI44Iq/perf.datao1Dpk6.2025032512120974",
]
```


src/run/runner/wall_time/perf.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+pub fn setup_environment() {
+    // Allow access to kernel symbols
+    run_with_sudo(&["sysctl", "-w", "kernel.kptr_restrict=0"]).unwrap();
+
+    // Allow non-root profiling
+    run_with_sudo(&["sysctl", "-w", "kernel.perf_event_paranoid=1"]).unwrap();
+}
```
art049: I think we might want to add this in the setup fn of the executor https://github.com/CodSpeedHQ/runner/blob/7b3a19c8fb4da9360d03fbbe1c188d2aa723856d/src/run/runner/executor.rs#L11
So this will automatically be ran by `codspeed setup`

art049: Also it would be interesting to check the value before setting it (so it will be possible to run `setup` as root but `run` without sudo)

art049 (commented):
A first quick review, I'll try to review more thouroughly

src/run/runner/wall_time/perf.rs
not-matthias: I moved the call to the `setup` function. I don't think we need to check for the value and it's not called in the `run` method anymore so it shouldn't need sudo.

art049: When you call run it still calls setup: https://github.com/CodSpeedHQ/runner/blob/7b3a19c8fb4da9360d03fbbe1c188d2aa723856d/src/run/mod.rs#L138

So in the same way we check valgrind is installed we could check beforehand if the system is configured without asking for sudo

not-matthias: Ahh, I see. Should be fixed now.

Cargo.toml
```diff
@@ -46,6 +46,8 @@ sysinfo = { version = "0.33.1", features = ["serde"] }
 indicatif = "0.17.8"
 console = "0.15.8"
 async-trait = "0.1.82"
+codspeed = { git = "https://github.com/CodspeedHQ/codspeed-rust", branch = "cod-674-collect-profiles-while-benchmarks-are-running", version = "=2.9.1" }
```
not-matthias: TODO: Replace with correct crate

src/run/runner/helpers/setup.rs
```diff
@@ -0,0 +1,33 @@
+use crate::prelude::*;
+use anyhow::{anyhow, bail, Result};
```
GuillaumeLagrange: All three are already included in the prelude

src/run/runner/wall_time/executor.rs
```diff
@@ -38,15 +44,80 @@ impl Executor for WallTimeExecutor {
             cmd.current_dir(abs_cwd);
         }
 
-        cmd.args(["-c", get_bench_command(config)?.as_str()]);
+        // Create the perf FIFOs before starting the command
+        let mut perf_fifo = PerfFifo::new();
+        let mut ctl_fifo = FifoIpc::create(RUNNER_CTL_FIFO)
+            .unwrap()
```
GuillaumeLagrange: Let's banish `unwrap()` usage altogether in favor of an anyhow error or at the very least an `expect` 😊

src/run/runner/helpers/run_command_with_log_pipe.rs
```diff
@@ -46,5 +49,13 @@ pub fn run_command_with_log_pipe(mut cmd: Command) -> Result<ExitStatus> {
     thread::spawn(move || {
         log_tee(stderr, std::io::stderr(), Some("[stderr]")).unwrap();
     });
+
+    let perf_pid = process.id();
```
GuillaumeLagrange: this is not necessarily a perf_pid here, this is just the argument command's pid

src/run/runner/helpers/run_command_with_log_pipe.rs
```diff
@@ -6,7 +6,10 @@ use std::process::Command;
 use std::process::ExitStatus;
 use std::thread;
 
-pub fn run_command_with_log_pipe(mut cmd: Command) -> Result<ExitStatus> {
+pub fn run_command_with_log_pipe_and_callback<F: FnOnce(u32)>(
+    mut cmd: Command,
+    cb: F,
```
GuillaumeLagrange, line 22: we could name this to indicate that the callback will get the process_id as argument

src/run/runner/wall_time/executor.rs
```diff
@@ -38,15 +44,80 @@ impl Executor for WallTimeExecutor {
...
+            .with_reader()
+            .unwrap();
 
+        // We have to pass a file to perf, which will create `perf.data.<timestamp>` files
+        // when the output is split.
+        let perf_dir = tempfile::tempdir()?;
```
GuillaumeLagrange: We may want to at least have an easy way to toggle whether we want to use perf or not, which can default to true

src/run/runner/wall_time/perf.rs
```diff
@@ -0,0 +1,63 @@
+use crate::run::runner::helpers::setup::run_with_sudo;
+use codspeed::fifo::FifoIpc;
```
GuillaumeLagrange: Have we given thought abouth whether the `FifoIPC` belong to `codpseed_rust` or to the codspeed runner ?

My initial impression is that it should be the runner that is the source of truth about what's hapenning in terms of FIFO, but this question may have an obvious answer once we tackle other languages than rust :thinking:

src/run/runner/wall_time/perf.rs
```diff
@@ -0,0 +1,63 @@
...
+use std::io::{Read, Write};
+
+pub const PERF_CTL_FIFO: &str = "/tmp/codspeed_perf.ctl.fifo";
```
GuillaumeLagrange: this should be an unique tmp dir for this runner instance imo, so that we don't have undefined behavior if people decide to run two walltime concurently (which is stupid, I give you that, but hey)

src/run/runner/wall_time/perf.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    pub fn new() -> Self {
+        // Note: The writer can't be opened before there's a reader. We can
+        // create our own reader first to be able to open the writer.
+        let ctl_fifo = FifoIpc::create(PERF_CTL_FIFO)
+            .unwrap()
+            .with_reader()
+            .unwrap()
+            .with_writer()
+            .unwrap();
```
GuillaumeLagrange: double unwrap

src/run/runner/wall_time/perf.rs
```diff
@@ -0,0 +1,63 @@
...
+        let ack_fifo = FifoIpc::create(PERF_CTL_ACK_FIFO)
+            .unwrap()
+            .with_reader()
+            .unwrap();
```
GuillaumeLagrange: double unwrap

GuillaumeLagrange (changes requested):
Good job overall ! Some minor details

src/run/runner/wall_time/perf.rs
```diff
@@ -0,0 +1,84 @@
...
+        self.wait_for_ack();
+    }
+
+    pub fn stop_events(&mut self) {
+        self.ctl_fifo.write_all(b"disable\n").unwrap();
+        self.wait_for_ack();
+    }
+}
+
+pub fn setup_environment() {
```
GuillaumeLagrange: Return a result instead of unwrapping here

src/run/runner/wall_time/executor.rs
```diff
@@ -1,24 +1,47 @@
...
+}
 
 #[async_trait(?Send)]
 impl Executor for WallTimeExecutor {
     fn name(&self) -> ExecutorName {
         ExecutorName::WallTime
     }
 
+    async fn setup(&self, _system_info: &SystemInfo) -> Result<()> {
+        super::perf::setup_environment();
```
GuillaumeLagrange: We could return the setup_environment result here, or at least `?` it to bubble up the error (and dont unwrap in the setup_environment)

src/run/runner/wall_time/perf.rs
```diff
@@ -55,8 +55,6 @@ impl PerfFifo {
 }
 
 pub fn setup_environment() {
-    // sysctl kernel.kptr_restrict=0
```
GuillaumeLagrange: I'm sure that's planned but you should fixup this to the previous  commit 😎

src/run/runner/wall_time/perf.rs
```diff
@@ -0,0 +1,84 @@
...
+pub const PERF_CTL_FIFO: &str = "/tmp/codspeed_perf.ctl.fifo";
+pub const PERF_CTL_ACK_FIFO: &str = "/tmp/codspeed_perf.ack.fifo";
+
+pub struct PerfFifo {
+    ctl_fifo: FifoIpc,
+    ack_fifo: FifoIpc,
+}
+
+impl PerfFifo {
+    fn wait_for_ack(&mut self) {
```
art049: This could be an  async function to avoid polling, and the use case matches well the purpose of the event loop

src/run/runner/wall_time/executor.rs
```diff
@@ -38,10 +60,65 @@ impl Executor for WallTimeExecutor {
...
+                    FifoCommand::StopBenchmark => {
+                        perf_fifo.stop_events();
+                        ack_fifo.send_cmd(FifoCommand::Ack).unwrap();
+                    }
+                    FifoCommand::Ack => unreachable!(),
+                }
+
+                std::thread::sleep(std::time::Duration::from_millis(10));
+            });
+        };
```
art049: in the same way, this could be an async task instead of a thread; it would avoid having to rely on polling

art049 (approved):
looking good overall :)

src/run/runner/wall_time/perf.rs
not-matthias: Yes! Planning to do fixup at the end so that I don't have a ton of force pushes before

src/run/runner/helpers/run_command_with_log_pipe.rs
not-matthias: Added a doc comment to explain the args, rather than changing the function name.

src/run/runner/wall_time/executor.rs
not-matthias: Added an environment variable `USE_PERF`. We can switch to CLI option later if needed. 

On second thought: Perf should only be enabled if it's supported by the integration, right? We don't have an option to switch for C++/Python/Rust projects atm. But since we start with perf disabled, we won't get any profiling data and it shouldn't be slowed down. So I think the env var is enough for now.

src/run/runner/wall_time/perf.rs
not-matthias: I'll copy the file into the runner once the codspeed-rust PR (https://github.com/CodSpeedHQ/codspeed-rust/pull/92) has been merged.

src/run/runner/mod.rs
```diff
@@ -5,7 +5,7 @@ use crate::prelude::*;
 use super::RunnerMode;
 
 mod executor;
-mod helpers;
+pub(crate) mod helpers;
```
GuillaumeLagrange: I don't think it needs to be public, at least build does not fail

src/run/runner/wall_time/executor.rs
```diff
@@ -1,24 +1,54 @@
...
 use async_trait::async_trait;
+use std::cell::OnceCell;
 use std::fs::canonicalize;
 use std::process::Command;
+use tempfile::TempDir;
+
+const PERF_DATA_PREFIX: &str = "perf.data.";
+
+pub struct WallTimeExecutor {
+    use_perf: bool,
```
GuillaumeLagrange: I find it a bit weird to have a `use_perf` bool, and two other fields that only make sense if `use_perf` is true.

Couldn't we encode this information better by wrapping all stuff that perf uses into a struct `Perf`, and just have an `Option<Perf>` that is simply `None` if we don't use perf?

This allows us to make impossible configuration simply not representable

src/run/runner/wall_time/executor.rs
```diff
@@ -1,24 +1,54 @@
...
+}
 
 #[async_trait(?Send)]
 impl Executor for WallTimeExecutor {
     fn name(&self) -> ExecutorName {
         ExecutorName::WallTime
     }
 
+    async fn setup(&self, _system_info: &SystemInfo) -> Result<()> {
+        super::perf::setup_environment()
```
GuillaumeLagrange: If we take for granted that we want to maintain a walltime executor that does not use perf:  
There's no need to call this if `perf` is not used.

src/run/runner/wall_time/executor.rs
```diff
@@ -38,12 +67,60 @@ impl Executor for WallTimeExecutor {
...
+        let status = if self.use_perf {
+            let perf_fifo = PerfFifo::new()?;
+            let runner_fifo = RunnerFifo::new()?;
+
+            // We have to pass a file to perf, which will create `perf.data.<timestamp>` files
+            // when the output is split.
+            let perf_file = tempfile::Builder::new()
+                .keep(true)
+                .prefix(PERF_DATA_PREFIX)
+                .tempfile_in(&self.perf_dir)?;
```
GuillaumeLagrange: could be a little `self.create_perf_file()` call

src/run/runner/wall_time/executor.rs
```diff
@@ -38,12 +67,60 @@ impl Executor for WallTimeExecutor {
...
+
+            cmd.args([
+                "-c",
+                &format!(
+                    "perf record --quiet --user-callchains --freq=max --switch-output --control=fifo:{},{} --delay=-1 -g --call-graph=dwarf --output={} -- {}",
```
GuillaumeLagrange: Shouldn't this belong to an helper function in the `perf` module? It feels a bit out of place here in the walltime executor doesn't it?

src/run/runner/wall_time/executor.rs
```diff
@@ -38,12 +67,60 @@ impl Executor for WallTimeExecutor {
...
+                    perf::fifo::handle_fifo(pid, runner_fifo, perf_fifo).await
+                });
+                thread_handle = Some(handle);
+
+                Ok(())
+            };
+            let status = run_command_with_log_pipe_and_callback(cmd, on_process_started);
+            info!("Process finished with status: {:?}", status);
+
+            // Write the bench_order to the perf directory
```
GuillaumeLagrange: This is a bit misleading as we are not writing here, we are just collecting bench order data from

src/run/runner/wall_time/perf/helpers.rs
```diff
@@ -0,0 +1,71 @@
+use std::collections::HashMap;
+
+use anyhow::{anyhow, bail};
+use linux_perf_data::{linux_perf_event_reader::EventRecord, PerfFileReader, PerfFileRecord};
+
+pub fn find_pid<P: AsRef<std::path::Path>>(path: P) -> anyhow::Result<i32> {
```
GuillaumeLagrange: 'll trust you on the fact that there's no simpler and more reliable way to find the pid ?

Also, either better argument name or small docstring here should be added to be a bit more precise on what pid we are looking for, and the fact that we are actually finding the pid from within the perf file.

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+mod shared;
+pub use shared::*;
+
+pub mod fifo;
+pub mod helpers;
+pub mod perf_map;
+pub mod unwind_data;
+
+pub struct BenchmarkData {
+    pub(crate) bench_order_by_pid: HashMap<u32, Vec<String>>,
```
GuillaumeLagrange: Could we use a better type from std to represent a PID rather than an `u32` ?

src/run/runner/wall_time/perf/perf_map.rs
```diff
@@ -0,0 +1,121 @@
+use log::warn;
```
GuillaumeLagrange: nitpick, this is in the crate prelude

src/run/runner/wall_time/perf/shared.rs
```diff
@@ -0,0 +1,41 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::Context;
+
+// !!!!!!!!!!!!!!!!!!!!!!!!
+// !! DO NOT TOUCH BELOW !!
```
GuillaumeLagrange: 👍

GuillaumeLagrange (changes requested):
Great piece of work! Some overall small changes, next should be the last round of review!

src/run/runner/wall_time/executor.rs
not-matthias: Yeah, good point. I'll create a separate struct for all the perf-related operations :+1:

src/run/runner/wall_time/perf/mod.rs
not-matthias: The std itself also uses u32 ([`std::process::id()`](https://doc.rust-lang.org/std/process/fn.id.html)). We could use `pid_t` from `libc` ([it isn't in the std](https://doc.rust-lang.org/std/os/unix/raw/type.pid_t.html) anymore). 

Alternatively, we could just wrap it in something like this: 
```rust
struct Pid(u32);
```

src/run/runner/wall_time/perf/helpers.rs
not-matthias: > 'll trust you on the fact that there's no simpler and more reliable way to find the pid ?

Sadly, yes. Perf never really samples a single process, so we can't get the pid from the filename or somewhere else. 
Getting the pid via the frequency, also allows us to detect potential issues with benchmarks overlapping or them being to short in the future. So it actually verifies our assumptions of how perf works. 

> Also, either better argument name or small docstring here should be added to be a bit more precise on what pid we are looking for, and the fact that we are actually finding the pid from within the perf file.

:+1:

src/run/runner/wall_time/perf/mod.rs
GuillaumeLagrange: I'm fine with aligning with std, thanks for the links!

src/run/runner/wall_time/perf/unwind_data.rs
```diff
@@ -0,0 +1,144 @@
+// !!!!!!!!!!!!!!!!!!!!!
+// !!! DO NOT MODIFY !!!
+// !!!!!!!!!!!!!!!!!!!!!
+//
```
art049: @not-matthias same here for the decoration, adding a warning below would be good enough

src/run/runner/wall_time/perf/shared.rs
```diff
@@ -0,0 +1,22 @@
...
+// !!!!!!!!!!!!!!!!!!!!!!!!
```
art049: @not-matthias same here for the decoration, adding a warning below would be good enough

src/main.rs
```diff
@@ -27,7 +27,7 @@ lazy_static! {
         format!("{VALGRIND_CODSPEED_BASE_VERSION}-0codspeed1");
 }
 
-#[tokio::main(flavor = "current_thread")]
+#[tokio::main(flavor = "multi_thread", worker_threads = 4)]
```
art049: Let's spend some time making sure this works with a single thread. Multithreading might fix but it's hiding another underlying issue

art049: Probably some blocking calls that could/should be non-blocking instead

src/run/runner/wall_time/executor.rs
```diff
@@ -38,12 +59,18 @@ impl Executor for WallTimeExecutor {
...
-        cmd.args(["-c", get_bench_command(config)?.as_str()]);
+        let bench_cmd = get_bench_command(config)?;
+        let status = if let Some(perf) = &self.perf {
+            perf.run(cmd, &bench_cmd).await
+        } else {
+            cmd.args(["-c", &bench_cmd]);
+            run_command_with_log_pipe(cmd)
+        };
 
-        debug!("cmd: {:?}", cmd);
```
GuillaumeLagrange, line 43: We're losing this debug log, which I think is not intended

src/run/runner/wall_time/perf/helpers.rs
```diff
@@ -0,0 +1,72 @@
...
+    let (pid, pid_count) = pid_freq
+        .iter()
+        .max_by_key(|&(_, count)| count)
+        .ok_or_else(|| anyhow!("Couldn't find pid in perf.data"))?;
+    log::debug!("Pid frequency: {:?}", pid_freq);
+
+    let pid_percentage = (*pid_count as f64 / total_count as f64) * 100.0;
+    if pid_percentage < 75.0 {
+        bail!(
+            "Pid {} has only {:.2}% of total events",
```
GuillaumeLagrange: `Most common pid` would make a clearer error message

src/run/runner/wall_time/perf/helpers.rs
```diff
@@ -0,0 +1,72 @@
...
+            i += 1;
+            if i >= COUNT_FIRST_N {
+                break;
+            }
+        }
+    }
+
+    // Choose the pid with the highest frequency. However, we can only use a pid if more than N% of the
+    // events are from that pid.
+    //
```
GuillaumeLagrange: Nitpick: extra line here not necessary

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -76,10 +79,21 @@ impl PerfRunner {
             .prefix(PERF_DATA_PREFIX)
             .tempfile_in(&self.perf_dir)?;
 
+        // Detect the mode based on the command to be executed
+        let cg_mode = if bench_cmd.contains("cargo") {
+            "dwarf"
+        } else if bench_cmd.contains("pytest") {
+            "fp"
+        } else {
+            warn!("Couldn't detect call graph mode for command: {}", bench_cmd);
```
GuillaumeLagrange: Shouldn't we throw rather than silently defaulting to dwarf ?

GuillaumeLagrange (approved):
Very minor comments appart from the multi-threading stuff we discussed IRL

src/run/runner/wall_time/perf/helpers.rs
not-matthias: The eternal struggle of mixing my personal comment style with other projects :sweat_smile: I'll fix it

src/run/runner/wall_time/perf/mod.rs
not-matthias: Yeah, good point. Changed it to a panic :+1:

src/main.rs
not-matthias: Yeah, good point. I remember the FIFO handler being an infinite loop which required it to be run on a separate thread. But it seems like the migration to `tokio::net::unix::pipe` simplified everything a bit and also allows single threaded runtime to be used. 

Works like a charm :+1:

not-matthias:
@GuillaumeLagrange Only minor changes in this PR: Added conditional dependency of `procfs`, disabled `USE_PERF` by default, install perf on setup if not found.

GuillaumeLagrange (approved):
lgtm, good work!

PR #74, opened 2025-03-26
Allow codspeed runner to run outside of a git directory

GuillaumeLagrange:
- Add `repository` and `provider` CLI arguments to `codspeed run`
- Add `--message-format=json` support to get created `runId` (only structured log for now)


src/logger.rs
```diff
@@ -71,3 +71,29 @@ pub(super) fn get_group_event(record: &log::Record) -> Option<GroupEvent> {
...
+        JSON_TARGET => {
+            let args = record.args().to_string();
+            if args.is_empty() {
+                None
+            } else {
+                Some(JsonEvent(args))
+            }
+        }
+        _ => None,
+    }
```
not-matthias: Can be converted to `if` since there's only one branch.

src/run/mod.rs
```diff
@@ -179,10 +204,32 @@ pub async fn run(args: RunArgs, api_client: &CodSpeedAPIClient) -> Result<()> {
...
         if provider.get_run_environment() == RunEnvironment::Local {
             start_group!("Fetching the results");
+            let run_id = upload_result.run_id.clone();
             poll_results::poll_results(api_client, &provider, upload_result.run_id).await?;
+            if output_json {
+                // TODO: Refactor `log_json` to accept a serde_json::Value arguemnt
+                log_json!(format!(
+                    "{{\"event\": \"run_finished\", \"run_id\": \"{}\"}}",
+                    run_id
+                ));
```
not-matthias, line 215: This seems very error-prone. I'd suggest rewriting this with [structured logging](https://docs.rs/log/latest/log/#structured-logging). You can then later get the values in the logger via [Records::key_values](https://docs.rs/log/latest/log/#structured-logging) (e.g. see [here](https://github.com/iorust/structured-logger/blob/main/src/lib.rs#L266-L268)). 

This would allow logging like this: 
```rust
log::info!(event: "run_finished"; run_id: run_id);
```

src/run/run_environment/local/provider.rs
```diff
@@ -17,85 +18,91 @@ use crate::run::{
...
+                    match config.repository_override.clone() {
+                        Some(RepositoryOverride {
+                            owner,
+                            repository,
+                            repository_provider,
+                        }) => (repository_provider, owner, repository),
+                        None => extract_provider_owner_and_repository_from_remote_url(
+                            remote.url().unwrap(),
+                        )?,
+                    };
```
not-matthias: `if let Some(...) = config.repository_override.clone() { } else {}` might be more suitable here

src/run/run_environment/local/provider.rs
```diff
@@ -17,85 +18,91 @@ use crate::run::{
...
+                    repository_root_path,
+                })
+            }
+            None => {
+                let RepositoryOverride {
+                    owner,
+                    repository,
+                    repository_provider,
+                } = config.repository_override.clone().context(
+                    "Could not find repository root and no respository was provided, \
```
not-matthias: typo: respository -> repository

not-matthias (approved):
Overall very nice, I'd only consider changing the `log_json` to use structured logging. Or maybe do it later instead of refactoring `log_json` to accept `serde_json::Value` (as mentioned in the comment)

src/run/mod.rs
not-matthias: This can also be combined with the already existing `log_json` to use a custom target.

GuillaumeLagrange: This is indeed much better, but I don't think it's worth to do it now.

I'll add this as a TODO next to json_log with a link to the `kv` feature.

Let's not overengineer something, there's a world where we do not use structured logging much more than just outputting the runId

src/run/mod.rs
```diff
@@ -55,6 +55,19 @@ pub struct RunArgs {
...
+    pub repository: Option<String>,
+
+    /// The repository provider to use in case --repository is used. Defaults to github
+    #[arg(
+        long,
+        env = "CODSPEED_PROVIDER",
+        requires = "repository",
+        ignore_case = true
+    )]
+    pub provider: Option<RepositoryProvider>,
```
adriencaccia, line 61: It would be best to hide those for now, as they will not be used by users at the moment

src/run/mod.rs
```diff
@@ -186,3 +201,17 @@ pub async fn run(args: RunArgs, api_client: &CodSpeedAPIClient) -> Result<()> {
 
     Ok(())
 }
+
+// We have to implement this manually, because deriving the trait makes the CLI values `git-hub`
+// and `git-lab`
+impl clap::ValueEnum for RepositoryProvider {
```
adriencaccia, line 226: Nice, I had the same issue with https://github.com/CodSpeedHQ/runner/blob/9829a0cb06da7bad3e4187232a11e2a9efa0ab65/src/run/mod.rs#L45, but decided to use `Walltime` instead of `WallTime` haha

src/run/run_environment/local/provider.rs
```diff
@@ -17,85 +18,91 @@ use crate::run::{
...
-    ))
-}
-
 impl TryFrom<&Config> for LocalProvider {
     type Error = Error;
-    fn try_from(_config: &Config) -> Result<Self> {
-        let repository_root_path = match find_repository_root(&std::env::current_dir()?) {
+    fn try_from(config: &Config) -> Result<Self> {
+        let current_dir = std::env::current_dir()?;
+        match find_repository_root(&current_dir) {
```
adriencaccia: This logic is a bit weird, if we are overriding the repository, we do not care about the actual repository state.

Let's simplify and branch directly on whether or not the repository is overridden, keeping the previous behavior is it is not.

adriencaccia: You will then be able to a test case for this, since there are no git repo repository involved

src/run/mod.rs
```diff
@@ -87,6 +87,9 @@ pub struct RunArgs {
     #[arg(long)]
     pub mongo_uri_env_name: Option<String>,
 
+    #[arg(long)]
+    pub message_format: Option<String>,
```
adriencaccia: We can have something a bit more explanatory, with an enum.
Hide it as well

adriencaccia (changes requested):

src/run/run_environment/local/provider.rs
GuillaumeLagrange: Are you sure we don't want to still get git info if it's available ? Like the branch and commit ref etc

The use case I am thinking about is when someone forks a repo, and wants to run the benches on the fork/upstream explicitely without relying on the `origin` remote

adriencaccia: Since we have not specified this behavior yet, I would prefer that we implement what we need at the moment.
Depending on what we end up specing for forks and local runs, the behavior you implemented is bound to change anyway.

src/run/mod.rs
GuillaumeLagrange: Yeah I also hesitated with this, it may be simpler down the line to not bother we implementing `ValueEnum` manually bit have better enum names

src/run/run_environment/local/provider.rs
```diff
@@ -31,78 +33,68 @@ impl TryFrom<&Config> for LocalProvider {
     type Error = Error;
     fn try_from(config: &Config) -> Result<Self> {
         let current_dir = std::env::current_dir()?;
-        match find_repository_root(&current_dir) {
+
+        if let Some(RepositoryOverride {
```
adriencaccia: Seen together, revert this logic to the previous behavior.
Keep the fake commit with the correct length though

src/local_logger.rs
```diff
@@ -105,6 +105,11 @@ impl Log for LocalLogger {
             return;
         }
 
+        if let Some(JsonEvent(json_string)) = get_json_event(record) {
+            println!("{json_string}");
+            return;
+        }
+
```
art049, line 112: I think we need to make sure those events are handled in the provider loggers to make sure we're not displaying this in CI environments

art049 (changes requested):

src/run/run_environment/local/provider.rs
GuillaumeLagrange: I made the logic a bit clearer by keeping the early return when not in a git repository, but reverted back

adriencaccia (approved):

art049 (approved):

PR #76, opened 2025-04-03
feat: output information about benches after a local run

GuillaumeLagrange:


src/api_client.rs
```diff
@@ -121,14 +124,26 @@ pub enum RunStatus {
...
     pub id: String,
-    pub status: RunStatus,
-    pub url: String,
-    pub head_reports: Vec<FetchLocalRunReportHeadReport>,
+    pub impact: Option<f64>,
+    pub conclusion: ReportConclusion,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct FetchLocalRunReportResult {
```
art049: this struct is not related to the report but the result only. This'd be closer to our internal data structures:
```suggestion
pub struct FetchLocalRunBenchmarkResult {
```

art049 (approved):

PR #79, opened 2025-04-23
feat: macro runner improvements

not-matthias:


not-matthias:
Assumes tokio is run with multiple worker threads, not compatible anymore. Will close this for now and document the designs on linear for later.

PR #81, opened 2025-05-14
Cod 870 implement landrun adapted runner commands

GuillaumeLagrange:
Needed to allow running commands in a sandbox with no network access

### DIFF
--- a/src/run/mod.rs
+++ b/src/run/mod.rs
@@ -9,6 +9,7 @@ use instruments::mongo_tracer::{install_mongodb_tracer, MongoTracer};
 use run_environment::interfaces::{RepositoryProvider, RunEnvironment};
 use runner::get_run_data;
 use serde::Serialize;
+use std::path::PathBuf;
 
 pub mod check_system;
 pub mod helpers;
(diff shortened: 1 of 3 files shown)


src/run/mod.rs
```diff
@@ -90,6 +95,15 @@ pub struct RunArgs {
         env = "CODSPEED_SKIP_UPLOAD"
     )]
     pub skip_upload: bool,
+    /// Used internally to upload the results after running the benchmarks in a sandbox environment
+    /// with no internet access
+    #[arg(
+        long,
+        default_value = "false",
+        hide = true,
+        env = "CODSPEED_SKIP_UPLOAD"
```
not-matthias: This should be `CODSPEED_UPLOAD_ONLY`

src/run/mod.rs
```diff
@@ -172,35 +188,40 @@ pub async fn run(
         end_group!();
     }
 
-    let run_data = get_run_data()?;
+    let run_data = get_run_data(&config)?;
 
-    start_opened_group!("Running the benchmarks");
+    if !(config.upload_only) {
```
not-matthias: unnecessary parentheses

not-matthias (approved):
It took me quite some time to realize that `upload_only` means. I think a better name for it would be `skip_run` since we already have `skip_setup` and `skip_upload`. And that would also make the purpose more obvious. 

Apart from the naming, everything else looks good! :+1:

src/run/mod.rs
GuillaumeLagrange: removed the env var as it makes no sense anyway here

PR #82, opened 2025-05-27
fix: valgrind crash for unresolved libpython

not-matthias:


src/run/runner/valgrind/measure.rs
```diff
@@ -73,11 +73,98 @@ fn create_run_script() -> anyhow::Result<TempPath> {
...
+    Ok(())
+}
+
 pub async fn measure(
     config: &Config,
     profile_folder: &Path,
     mongo_tracer: &Option<MongoTracer>,
 ) -> Result<()> {
+    // Ensure that valgrind can find the libpython library when using uv:
+    if config.command.contains("uv run") {
```
art049: I think this would still appear even if you activate the env before:
```
uv venv
source .venv/bin/activate
```
and then codspeed run python ...
Can you test this? This would mean that this check is not enough

not-matthias: Ohh, true:
```
(reflex-perf) [11:20] not-matthias:reflex-perf (main)> codspeed run -- pytest tests/benchmarks/                                                                                                  devenv-shell-env
Unofficially supported system: nixos 25.05. Continuing with best effort support.

►►► Preparing the environment
Valgrind is already installed with the correct version, skipping installation
Environment ready

►►► Running the benchmarks
/home/not-matthias/Documents/work/wgit/reflex-perf/.venv/bin/python3: error while loading shared libraries: /home/not-matthias/Documents/work/wgit/reflex-perf/.venv/bin/../lib/libpython3.13.so.1.0: cannot open shared object file: No such file or directory
Error: failed to execute the benchmark process, exit code: 127
```

I guess we should check for `config.command.contains("pytest")` and check whether `uv` is installed

art049: I think we can go even further and assume you'll be using Python from the path directly (basically the same assumption we have to find the objects to exclude already [here](https://github.com/CodSpeedHQ/runner/blob/d3c5540a02cbb1efa6e569723144bace4fcb5426/src/run/runner/valgrind/helpers/ignored_objects_path.rs#L4)).

So all the time, if we find out the shared library is improperly linked, we can print a warning and try to fix it (and print an error without breaking anything if we don't manage to do it)

art049 (commented):
That's nice, yes. It seems better to have this logic in a shell script. Can you include it using `include!` so we can benefit from syntax highlighting and linting (I use shellcheck locally, for example. We probably don't need to include this lint in CI, though.

src/run/runner/valgrind/helpers/venv_compat.sh
```diff
@@ -0,0 +1,62 @@
...
+        return 1
+    fi
+
+    system_path="$(dirname $(dirname "$system_python"))"
+    venv_path="$(dirname $(dirname "$venv_python"))"
+    echo "Python installation (system): $system_path"
+    echo "Python installation (venv): $venv_path"
+
+    # Find libpython in the system python dir
+    mapfile -t libpython_array < <(find "$system_path" -maxdepth 3 -type f -name "libpython*.so.1.0" 2>/dev/null)
```
art049: Not sure we'll always have the `.1.0` suffix could be worth it to add a wildcard as well if it doesn't break anything

src/run/runner/valgrind/helpers/venv_compat.sh
```diff
@@ -0,0 +1,62 @@
...
+uv_python="$(uv python find 2>/dev/null || true)"
+if [ -n "$uv_python" ]; then
+    add_symlink "$uv_python"
+else
+    echo "Didn't find uv venv, continuing..."
+fi
+
+##
+##
+##
```
art049: I guess we can remove the `#` 
```suggestion
uv_python="$(uv python find 2>/dev/null || true)"
if [ -n "$uv_python" ]; then
    add_symlink "$uv_python"
else
    echo "Didn't find uv venv, continuing..."
fi
```

art049 (approved):
looks good

PR #86, opened 2025-06-04
feat: add cpu isolation for system, perf and benchmark

not-matthias:


src/run/runner/helpers/isolation.rs
```diff
@@ -0,0 +1,63 @@
...
+        Ok(())
+    }
+
+    pub fn isolate_user(&self, allowed_cpus: &[u32]) -> anyhow::Result<()> {
+        Self::set_allowed_cpus("user.slice", allowed_cpus)?;
+
+        Ok(())
+    }
+
+    pub fn reset_isolation(nproc: u32) -> anyhow::Result<()> {
```
GuillaumeLagrange: There should be no argument, and we should take nproc from self to make sure one can't mess up this call

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -96,10 +97,21 @@ impl PerfRunner {
         };
         debug!("Using call graph mode: {}", cg_mode);
 
+        const PERF_CPU: u32 = 2;
+        const BENCH_CPU: u32 = 3;
+        const BENCH_CPUS: [u32; 2] = [PERF_CPU, BENCH_CPU];
+        const HOST_CPUS: [u32; 2] = [0, 1];
```
GuillaumeLagrange: can you add comments to explain what is going on here ?

especially why we have two host CPUs for example.

GuillaumeLagrange (changes requested):
Two subsequent questions:

1. Did you try the effect of existing walltime benchmarks ? We may need to prepare users for breaking changes on this regard.

2. How does this behave when the user is not root on the machine ? (Edge case obviously since macro runners are not concerned)

src/run/runner/wall_time/perf/mod.rs
not-matthias: Will do! So my main thoughts behind this was to use core 0 for system, but core 1 is often on the same physical core. So I wanted to use the next pair available. But I just realized we could put the perf process on that core since we don't really care about that performance. 

I think the following makes more sense: 
- Core 0: Reserved for the system
- Core 1: Reserved for perf
- Core 2: Reserved for the benchmark

not-matthias: Since we're adding the `isolcpus` parameter to the runner images, we should use the isolated CPUs by checking `/sys/devices/system/cpu/isolated`. Added a fallback to use the normal CPUs. 

EDIT: Turns out the perf process must also be usable from the system.slice
EDIT: Works as expected: 
```
[DEBUG::codspeed::run::runner::wall_time::perf] Using call graph mode: dwarf
[DEBUG::codspeed::run::runner::helpers::isolation] Isolated CPUs: [10, 14, 15]
[DEBUG::codspeed::run::runner::helpers::isolation] Root CPUs: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12, 13]
[DEBUG::codspeed::run::runner::helpers::isolation] Root CPUs: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12, 13]
[DEBUG::codspeed::run::runner::helpers::isolation] Isolated CPUs: [10, 14, 15]
[DEBUG::codspeed::run::runner::helpers::isolation] Root CPUs: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12, 13]
[DEBUG::codspeed::run::runner::helpers::isolation] Root CPUs: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12, 13]
[DEBUG::codspeed::run::runner::helpers::setup] Running command: sudo systemctl set-property system.slice AllowedCPUs=0,1,2,3,4,5,6,7,8,9,11,12,13
[DEBUG::codspeed::run::runner::helpers::setup] Running command: sudo systemctl set-property init.scope AllowedCPUs=0,1,2,3,4,5,6,7,8,9,11,12,13
[DEBUG::codspeed::run::runner::helpers::setup] Running command: sudo systemctl set-property user.slice AllowedCPUs=0,10
[DEBUG::codspeed::run::runner::wall_time::perf] cmd: ARCH="x86_64" CODSPEED_ENV="runner" CODSPEED_PROFILE_FOLDER="/tmp/profile.usHSTs0vY0.out" CODSPEED_RUNNER_MODE="walltime" PYTHONHASHSEED="0" "bash" "-c" "taskset -c 0 perf record --user-callchains --freq=999 --switch-output --control=fifo:/tmp/.tmpJGb7DK/codspeed_perf.ctl.fifo,/tmp/.tmpJGb7DK/codspeed_perf.ack.fifo --delay=-1 -g --call-graph=dwarf --output=/tmp/.tmpO06ohj/perf.data.4GKjT9 -- taskset -c 10 cargo-codspeed run"
Events disabled
[cargo-codspeed] Measurement mode: Walltime

Collected 1 benchmark suite(s) to run
Running perf-walltime-repo my_benchmark
Timer precision: 45 ns
my_benchmark  fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ fibonacci                │               │               │               │         │
```

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -96,13 +96,16 @@ impl PerfRunner {
         };
         debug!("Using call graph mode: {}", cg_mode);
 
+        let user = nix::unistd::User::from_uid(nix::unistd::Uid::current())?.unwrap();
         cmd.args([
             "-c",
             &format!(
-                "perf record --user-callchains --freq=999 --switch-output --control=fifo:{},{} --delay=-1 -g --call-graph={cg_mode} --output={} -- {bench_cmd}",
+                "perf record --user-callchains --freq=999 --switch-output --control=fifo:{},{} --delay=-1 -g --call-graph={cg_mode} --output={} -- \
+                sudo systemd-run --scope --slice=codspeed.slice --same-dir -- runuser -u {} -- {bench_cmd}",
```
art049, line 116: will this work if the slice doesn't exist? we should probably check the existence

not-matthias: yes, it will just create a new slice with this name under `system.slice`

art049: ok

art049 (approved):

PR #87, opened 2025-06-10
fix: only show perf output at debug or trace level

not-matthias:


not-matthias:
> can't we redirect the output to a log file we could send in the archive instead? I think we're doing this with valgrind already:

See Linear issue: Unfortunately perf-record doesn't allow writing the outputs to a file so I decided to disable it when it's not running at a DEBUG or TRACE log level.

art049 (approved):

PR #89, opened 2025-06-17
Cod 1008 benchmark run fails on macro runner with systemd run

not-matthias:
**Fix 1:** Turns out that runuser doesn't work without `sudo` in the macro-runner: 
```
runner@ip-172-31-0-162:~$ runuser -u runner ls
runuser: may not be used by non-root users
runner@ip-172-31-0-162:~$ sudo runuser -u runner whoami
runner
```

I decided to use the `uid` and `gid` parameters in `systemd-run` since we don't have to rely on runuser in that case. 

**Fix 2:** We also have to pass the environment variables to the process, since the benchmarks are running with systemd. 

**Fix 3**: Perf can't trace a process that's running with root privileges. We have to run perf with root, which makes the child processes also run with root. I also updated the splitting, since we can't kill a root process, from a non-root process. 

EDIT: @art049 should've waited with the merge, the last 3 commits aren't related to this PR (they are from the other one that was stacked on top of this one). Please ignore them since the review was already done

### DIFF
--- a/src/run/runner/wall_time/executor.rs
+++ b/src/run/runner/wall_time/executor.rs
@@ -1,11 +1,12 @@
 use super::perf::PerfRunner;
 use crate::prelude::*;
+use crate::run::RunnerMode;
 use crate::run::instruments::mongo_tracer::MongoTracer;
 use crate::run::runner::executor::Executor;
 use crate::run::runner::helpers::env::get_base_injected_env;
 use crate::run::runner::helpers::get_bench_command::get_bench_command;
 use crate::run::runner::helpers::run_command_with_log_pipe::run_command_with_log_pipe;
-use crate::run::runner::{ExecutorName, RunData, RunnerMode};
+use crate::run::runner::{ExecutorName, RunData};
 use crate::run::{check_system::SystemInfo, config::Config};
 use async_trait::async_trait;
 use std::fs::canonicalize;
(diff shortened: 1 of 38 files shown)


src/run/runner/wall_time/executor.rs
```diff
@@ -44,25 +57,18 @@ impl Executor for WallTimeExecutor {
...
             cmd.current_dir(abs_cwd);
         }
 
-        let bench_cmd = get_bench_command(config)?;
+        let bench_cmd = Self::walltime_bench_cmd(config, run_data)?;
         let status = match (config.enable_perf, &self.perf) {
             (true, Some(perf)) => perf.run(cmd, &bench_cmd, config).await,
             _ => {
-                cmd.args(["-c", &bench_cmd]);
+                cmd.args(["-c", &format!("sudo {bench_cmd}")]);
```
art049: maybe it would be better to handle sudo in the top level command instead

src/run/runner/wall_time/executor.rs
```diff
@@ -82,13 +88,13 @@ impl Executor for WallTimeExecutor {
...
         &self,
-        _config: &Config,
+        config: &Config,
         _system_info: &SystemInfo,
         run_data: &RunData,
     ) -> Result<()> {
         debug!("Copying files to the profile folder");
 
-        if let Some(perf) = &self.perf {
+        if let (Some(perf), true) = (&self.perf, config.enable_perf) {
```
art049: Seems the perfect use case for if let chains released in 1.88! https://rust-lang.github.io/rfcs/2497-if-let-chains.html

art049 (approved):

src/run/runner/wall_time/executor.rs
art049: Mostly, if we want to allow piping stdin to enter the password. It would also clarify where we need root permissions

not-matthias: We're now spawning a `sudo` process and pass `sh -c` as args.

src/run/runner/wall_time/executor.rs
not-matthias: Had to update update the rust-toolchain to 1.88 and the edition to 2024.

art049 (approved):

PR #91, opened 2025-06-20
feat: enable valgrind fair-sched

CPunisher:
From https://valgrind.org/docs/manual/manual-core.html#manual-core.pthreads:

> The fairness of the futex based locking produces better reproducibility of thread scheduling for different executions of a multithreaded application. 

I'm not sure whether you have tried or this is helpful to codspeed.


CPunisher:
Noted that `lai-callgrind` enable it by default: https://iai-callgrind.github.io/iai-callgrind/latest/html/benchmarks/library_benchmarks/important.html

PR #93, opened 2025-06-24
feat: add pre- and post-benchmark scripts

not-matthias:
This allows us to run custom logic on the macro runner (e.g. set core isolation) without having all of that logic inside the runner.

### DIFF
--- a/src/run/runner/valgrind/helpers/venv_compat.rs
+++ b/src/run/runner/valgrind/helpers/venv_compat.rs
@@ -51,6 +51,8 @@ pub fn symlink_libpython(cwd: Option<&String>) -> anyhow::Result<()> {
 mod tests {
     use super::*;
 
+    // Only run in Github Actions, to ensure python is dynamically linked.
+    #[test_with::env(GITHUB_ACTIONS)]
     #[test]
     fn test_venv_compat_no_crash() {
         assert!(symlink_libpython(None).is_ok());
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -312,6 +312,7 @@ dependencies = [
  "sysinfo",
  "temp-env",
  "tempfile",
+ "test-with",
  "tokio",
  "tokio-tar",
  "url",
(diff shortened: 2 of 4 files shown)


src/run/runner/valgrind/helpers/venv_compat.rs
```diff
@@ -53,6 +53,6 @@ mod tests {
 
     #[test]
     fn test_venv_compat_no_crash() {
-        assert!(symlink_libpython(None).is_ok());
+        let _ = symlink_libpython(None);
     }
 }
```
art049: Uh, did I miss something?

not-matthias: Ahh, yeah. Probably should've created a linear issue/separate PR for this. On my system this test failed because libpython is statically linked. I've created a separate commit and wanted to include it in this PR, but I guess it's better to create a new one.

art049: Ah, okay, well, we still need to keep this test, though

not-matthias: > Ah, okay, well, we still need to keep this test, though

Yeah, definitely. But we can't really check if we patched it since there's no way to ensure that libpython is dynamically linked. So my intention here was to just check if it ran, and didn't crash.

art049: maybe we can have something to skip it locally and run it still in the CI?
https://nexte.st/docs/running/

art049: Or https://crates.io/crates/test-with only when `CI` is set

not-matthias: Fixed it. We're now only running it in CI using the `test-with` create (nice find!). Is it fine to keep it in this PR, feels a bit overkill to create a new one/ticket for this.

art049 (approved):
lgtm!

src/run/runner/valgrind/helpers/venv_compat.rs
art049: yep thats fine

PR #94, opened 2025-06-25
feat: improve callgraph mode detection

not-matthias:


src/run/runner/wall_time/executor.rs
```diff
@@ -18,7 +18,9 @@ pub struct WallTimeExecutor {
 impl WallTimeExecutor {
     pub fn new() -> Self {
         let use_perf = if cfg!(target_os = "linux") {
-            std::env::var("CODSPEED_USE_PERF").is_ok()
+            std::env::var("CODSPEED_PERF_ENABLED")
```
GuillaumeLagrange: let's move this to a util function that takes the env var name as argument

and we might as well go all in on truthy values

`"1", "true", "yes", "on", "enabled"`
I don't have others in mind

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -84,15 +84,17 @@ impl PerfRunner {
...
-            "fp"
+            "fp".into()
         } else {
-            panic!(
-                "Perf not supported. Failed to detect call graph mode for command: {}",
-                bench_cmd
-            )
+            // Default to dwarf unwinding since it works well with most binaries.
+            debug!("No call graph mode detected, defaulting to dwarf");
+            "dwarf".into()
```
GuillaumeLagrange:
```suggestion
let cg_mode = match (std::env::var("CODSPEED_PERF_CALLGRAPH_MODE").ok(), &bench_cmd) {
    (Some(mode), _) => {
        debug!("Using call graph mode from environment variable: {}", mode);
        mode
    }
    (None, cmd) if cmd.contains("pytest") => "fp".into(),
    (None, cmd) if cmd.contains("cargo") => "dwarf".into(),
    (None, _) => {
        // Default to dwarf unwinding since it works well with most binaries.
        debug!("No call graph mode detected, defaulting to dwarf");
        "dwarf".into()
    }
};
```

wdyt ?

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -84,15 +84,17 @@ impl PerfRunner {
             .tempfile_in(&self.perf_dir)?;
```
GuillaumeLagrange, line 91: nitpick: commit log is weird, I don't really understand the sentence without context + random semicolumn

GuillaumeLagrange (changes requested):
olgtm, second round will be quick!

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -84,15 +84,17 @@ impl PerfRunner {
...
 
         // Detect the mode based on the command to be executed
-        let cg_mode = if bench_cmd.contains("cargo") {
-            "dwarf"
+        let cg_mode = if let Ok(mode) = std::env::var("CODSPEED_PERF_CALLGRAPH_MODE") {
```
art049: We should add this as a clap CLI flag (maybe still allow the env variable though). The less we rely on env the better. 
Similarly enabling perf profiling could be in a CLI flag as well.

src/run/runner/wall_time/perf/mod.rs
art049: Also it seems that we could abstract this callgraph mode option name. For perf, this is named callgraph mode, but for us it's how we resolve symbols, no? We should try to decouple it so we can use this flag with other tools if needed

src/run/runner/wall_time/perf/mod.rs
art049: This might also fix what @GuillaumeLagrange mentionned on the truth values since I guess clap has something built in for this

art049 (changes requested):

src/run/runner/wall_time/perf/mod.rs
not-matthias: The callgraph mode is mostly about how we do the unwinding to find the callstack. So we could rename it to `CODSPEED_UNWINDING_MODE`, although we can only set it for certain tools (e.g. perf). 

For bpftrace, we don't have the option to support `fp` unwinding, so this could maybe be confusing :thinking: Valgrind also uses DWARF by default, and doesn't allow switching to `fp` unwinding. 

So I'd propose keeping the name related to perf (e.g. `PERF_UNWINDING_MODE`) as we can easily rename it later on when we need it.

not-matthias:
Alright, fixed all the comments. Only slight downside of having the `enable_perf` flag in the CLI arguments is that it's not possible to install the perf binary conditionally. This is because the `setup` command doesn't include the same flags so we're currently always installing perf (which shouldn't be an issue since at some point we're going to switch to this anyways).

src/run/runner/wall_time/perf/mod.rs
art049: ok make sense

src/run/mod.rs
```diff
@@ -38,6 +38,29 @@ fn show_banner() {
...
+#[derive(Args, Debug, Clone)]
+pub struct PerfRunArgs {
+    /// Enable the performance runner, which uses `perf` to collect performance data.
+    /// This is only supported on Linux.
+    #[arg(long, env = "CODSPEED_PERF_ENABLED", default_value_t = false)]
+    enable_perf: bool,
+
+    /// The unwinding mode that should be used with perf to collect the call stack.
+    #[arg(long, env = "CODSPEED_PERF_UNWINDING_MODE")]
+    unwinding_mode: Option<UnwindingMode>,
```
art049: from your comment then, making this `--perf-unwinding-mode` would make more sense so you can either just rename the CLI flag or the variable !

art049 (approved):

PR #95, opened 2025-07-07
fix: runner doesnt work with in GitHub action

not-matthias:
Fixed the valgrind safe-run (missing quotation). Added tests for the walltime and valgrind executors. We're now also forwarding the system environment variables in the `systemd-run` cmd.

### DIFF
--- a/src/run/runner/tests.rs
+++ b/src/run/runner/tests.rs
@@ -1,0 +1,187 @@
+use crate::run::check_system::SystemInfo;
+use crate::run::config::Config;
+use crate::run::runner::executor::Executor;
+use crate::run::runner::interfaces::RunData;
+use crate::run::runner::valgrind::executor::ValgrindExecutor;
+use crate::run::{RunnerMode, runner::wall_time::executor::WallTimeExecutor};
+use tempfile::TempDir;
+use tokio::sync::{OnceCell, Semaphore, SemaphorePermit};
+
+const SIMPLE_ECHO_SCRIPT: &str = "echo 'Hello, World!'";
+const MULTILINE_ECHO_SCRIPT: &str = "echo \"Working\"
+echo \"with\"
+echo \"multiple lines\"";
+const MULTILINE_ECHO_WITH_SEMICOLONS: &str = "echo \"Working\";
+echo \"with\";
+echo \"multiple lines\";";
+const DIRECTORY_CHECK_SCRIPT: &str = "cd /tmp
+# Check that the directory is actually changed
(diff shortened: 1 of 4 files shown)


src/run/runner/wall_time/executor.rs
```diff
@@ -26,6 +26,11 @@ impl WallTimeExecutor {
     fn walltime_bench_cmd(config: &Config, run_data: &RunData) -> Result<String> {
         let bench_cmd = get_bench_command(config)?;
 
+        // We have to forward all the environment variables.
```
GuillaumeLagrange: I fear that we'll end up with unwanted information in the command, which is written to our runner.log and uploaded on our servers. Am I overthinking this ?

src/run/runner/tests.rs
art049, line 1: Nice tests, I think it would make sense to parameterize them / use a fixture instead of iterating: https://github.com/la10736/rstest

GuillaumeLagrange:
+ ci is failing

art049 (approved):
Only feedback is on the test strucure but mostly cosmetic

src/run/runner/wall_time/executor.rs
not-matthias: Good point, we could end up logging secrets (e.g. `GITHUB_TOKEN`). Unfortunately I don't see an easier way to do this. 

The only other way I see right now is to add a CLI arg to forward specific variables, but this would be a breaking change for users who use env vars already. 

Can we iterate over all of the env vars in Github Actions and then forward them manually from our action? Does this also work for Gitlab and Buildkite?

art049: Systemd has EnvironmentFile so adding `--property=EnvironmentFile={env_file_path}` should work
[docs](https://www.freedesktop.org/software/systemd/man/latest/systemd-run.html#--property=)

src/run/runner/wall_time/executor.rs
```diff
@@ -23,19 +24,35 @@ impl WallTimeExecutor {
...
+        let system_env = std::env::vars()
+            .map(|(env, value)| format!("{env}={value}"))
+            .join("\n");
+
+        let base_env = get_base_injected_env(RunnerMode::Walltime, &run_data.profile_folder)
             .into_iter()
-            .map(|(env, value)| format!("--setenv={env}={value}"))
-            .join(" ");
+            .map(|(env, value)| format!("{env}={value}"))
+            .join("\n");
```
art049: Since we use the same formatting, we could chain the iterators here

PR #96, opened 2025-07-09
fix: execute pre- and post-bench scripts for non-perf walltime runner

not-matthias:
- Moved it to the walltime runner
- Changed the log warnings to debug

### DIFF
--- a/src/run/runner/wall_time/executor.rs
+++ b/src/run/runner/wall_time/executor.rs
@@ -11,8 +11,57 @@ use crate::run::{check_system::SystemInfo, config::Config};
 use async_trait::async_trait;
 use std::fs::canonicalize;
 use std::io::Write;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use tempfile::NamedTempFile;
+
+struct HookScriptsGuard {
+    post_bench_script: PathBuf,
+}
+
+impl HookScriptsGuard {
+    fn execute_script_from_path<P: AsRef<Path>>(path: P) -> anyhow::Result<()> {
+        let path = path.as_ref();
+        if !path.exists() || !path.is_file() {
+            debug!("Script not found or not a file: {}", path.display());
+            return Ok(());
+        }
+
+        let output = Command::new("bash").args([&path]).output()?;
+        if !output.status.success() {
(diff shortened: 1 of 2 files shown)


src/run/runner/wall_time/executor.rs
```diff
@@ -11,9 +11,58 @@ use crate::run::{check_system::SystemInfo, config::Config};
...
+            debug!("stderr: {}", String::from_utf8_lossy(&output.stderr));
+            bail!("Failed to execute script: {}", path.display());
+        }
+
+        Ok(())
+    }
+
+    pub fn setup_with_scripts<P: AsRef<Path>>(pre_bench_script: P, post_bench_script: P) -> Self {
+        if let Err(e) = Self::execute_script_from_path(pre_bench_script.as_ref()) {
+            warn!("Failed to execute pre-bench script: {e}");
```
art049: that means everytime we will run with the walltime executor on a non walltime runner, we'll encounter this warning. This should probably a debug log instead.

src/run/runner/wall_time/executor.rs
```diff
@@ -11,9 +11,58 @@ use crate::run::{check_system::SystemInfo, config::Config};
...
+            "/usr/local/bin/codspeed-pre-bench",
+            "/usr/local/bin/codspeed-post-bench",
+        )
+    }
+}
+
+impl Drop for EnvGuard {
+    fn drop(&mut self) {
+        if let Err(e) = Self::execute_script_from_path(&self.post_bench_script) {
+            warn!("Failed to execute post-bench script: {e}");
```
art049: same here

src/run/runner/wall_time/executor.rs
```diff
@@ -11,9 +11,58 @@ use crate::run::{check_system::SystemInfo, config::Config};
...
 use std::io::Write;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use tempfile::NamedTempFile;
 
+struct EnvGuard {
+    post_bench_script: PathBuf,
+}
+
+impl EnvGuard {
```
art049: not related to this PR but this could use a more explicit name like `HookScriptsGuard`

art049 (approved):

PR #97, opened 2025-07-09
fix: run with --scope to allow perf to trace the benchmark process

not-matthias:
Unfortunately when running with --scope, we can't use `EnvironmentFile` which is why I had to add it to the sh cmd. Perf can now trace the benchmark process again.

### DIFF
--- a/src/run/runner/wall_time/perf/perf_map.rs
+++ b/src/run/runner/wall_time/perf/perf_map.rs
@@ -51,6 +51,20 @@ impl ModuleSymbols {
             return Err(anyhow::anyhow!("No symbols found"));
         }
 
+        // The base_addr from the mapping is where the module is actually loaded in memory (See ProcessSymbols::add_mapping),
+        // but the symbol addresses from the ELF file assume the module is loaded at its preferred virtual address. We need to:
+        // 1. Find the module's preferred base address from the ELF file or symbols
+        // 2. Calculate the offset: actual_base - preferred_base
+        // 3. Apply this offset to the symbol addresses
+
+        // Find the preferred base address from the minimum symbol address
+        let preferred_base = symbols.iter().map(|s| s.offset).min().unwrap_or(0) & !0xfff; // Align to page boundary
+
+        // Convert absolute addresses to relative offsets
+        for symbol in &mut symbols {
+            symbol.offset = symbol.offset.saturating_sub(preferred_base);
+        }
+
         Ok(Self {
             path: path.as_ref().to_path_buf(),
             symbols,


GuillaumeLagrange (changes requested):
--scope actually inherits env from what I have see

```
export MY_VAR="hello"

systemd-run env | grep MY_VAR

systemd-run --scope env | grep MY_VAR
```

not-matthias:
> --scope actually inherits env from what I have see
> 
> ```
> export MY_VAR="hello"
> 
> systemd-run env | grep MY_VAR
> 
> systemd-run --scope env | grep MY_VAR
> ```

Doesn't work on my machine. Seems to only inherit the system env: 
```bash
> export THIS_IS_A_TEST="ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ"
> echo $THIS_IS_A_TEST
ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
> sudo systemd-run --scope -- env | grep THIS_IS_A_TEST
Running as unit: run-p2129159-i2129160.scope; invocation ID: fdf21a15d503409891f434dd43243ad1
```

src/run/runner/wall_time/executor.rs
```diff
@@ -34,7 +34,11 @@ impl WallTimeExecutor {
                     .into_iter()
                     .map(|(k, v)| (k.into(), v)),
             )
-            .map(|(env, value)| format!("{env}={value}"))
+            .map(|(env, value)| {
+                // Properly quote the value to handle special characters
+                let escaped_value = value.replace("\\", "\\\\").replace("\"", "\\\"");
+                format!("export {env}=\"{escaped_value}\"")
```
art049: ouch creating a file we need to source like this seems like we can introduce a ton of annoying edge cases

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -251,7 +251,10 @@ impl PerfRunner {
         let perf_map_pids = futures::future::try_join_all(copy_tasks)
             .await?
             .into_iter()
-            .filter_map(Result::ok)
+            .filter_map(|result| {
+                debug!("Copy task result: {result:?}");
+                result.ok()
+            })
```
art049, line 204: should we keep this?

art049 (changes requested):
Is using source the only way to inherit the env?

src/run/runner/wall_time/perf/mod.rs
not-matthias: It helped me figure out why the perf maps weren't copied to the profile folder. I think it makes sense to keep this since otherwise we have no idea what's going on. It's just a debug log so we won't see it 99% of the time anyway.

src/run/runner/wall_time/executor.rs
not-matthias: So the only other way that I found is writing `{env}={value}` to the file and then loading it with `export (cat path_to_env | xattr)`. I like the `export {env}={value}` approach more since it's simpler. 

However, we're now using the `shell-escape` crate to simplify the escaping.

not-matthias: We could also use the `printenv` /  [`export` cmd](https://unix.stackexchange.com/a/435321) and write it to the file directly.

not-matthias: We're now using the `export` cmd

src/run/runner/tests.rs
```diff
@@ -86,18 +136,23 @@ mod valgrind {
...
-    #[case("MY_ENV_VAR", "Hello", ENV_VAR_VALIDATION_SCRIPT)]
+    #[case(ENV_TESTS[0])]
+    #[case(ENV_TESTS[1])]
+    #[case(ENV_TESTS[2])]
+    #[case(ENV_TESTS[3])]
+    #[case(ENV_TESTS[4])]
+    #[case(ENV_TESTS[5])]
+    #[case(ENV_TESTS[6])]
+    #[case(ENV_TESTS[7])]
     #[tokio::test]
```
art049: For readability purpose, we should probably use something like rstest_reuse or find another pattern 
https://docs.rs/rstest/latest/rstest/attr.rstest.html#use-parametrize-definition-in-more-tests

src/run/runner/tests.rs
```diff
@@ -163,20 +222,34 @@ mod walltime {
...
+    #[case(ENV_TESTS[3], false)]
+    #[case(ENV_TESTS[3], true)]
+    #[case(ENV_TESTS[4], false)]
+    #[case(ENV_TESTS[4], true)]
+    #[case(ENV_TESTS[5], false)]
+    #[case(ENV_TESTS[5], true)]
+    #[case(ENV_TESTS[6], false)]
+    #[case(ENV_TESTS[6], true)]
+    #[case(ENV_TESTS[7], false)]
+    #[case(ENV_TESTS[7], true)]
```
art049: you can parametrize enable_perf independently here so it will do a cartesian product of the parameter sets and improve readibility
using something like  #[values(true, "false)] on top of the rstest reuse

Cargo.toml
```diff
@@ -54,6 +54,7 @@ debugid = "0.8.0"
 memmap2 = "0.9.5"
 nix = { version = "0.29.0", features = ["fs", "user"] }
 futures = "0.3.31"
+shell-quote = "0.7.2"
```
art049: Can you explain a bit why we need this library?

src/run/runner/tests.rs
```diff
@@ -19,19 +20,66 @@ const DIRECTORY_CHECK_SCRIPT: &str = "cd /tmp
...
+
+const ENV_TESTS: [(&str, &str); 8] = [
+    ("quotes_and_escapes", QUOTES_AND_ESCAPES),
+    ("multiline_and_whitespace", MULTILINE_AND_WHITESPACE),
+    ("shell_metacharacters", SHELL_METACHARACTERS),
+    ("variables_and_commands", VARIABLES_AND_COMMANDS),
+    ("unicode_and_special", UNICODE_AND_SPECIAL),
+    ("complex_mixed", COMPLEX_MIXED),
+    ("empty", EMPTY),
+    ("space_only", SPACE_ONLY),
```
art049: also defining the tests all in the same const would help readability of the test cases

art049 (changes requested):

Cargo.toml
not-matthias: I was initially using this for escaping the runner args, but now we are only using it to create the script to validate the env var: 
```rust
fn env_var_validation_script(env: &str, expected: &str) -> String {
    let expected: String = expected.quoted(Bash);
    format!(
        r#"
if [ "${env}" != {expected} ]; then
  echo "FAIL: Environment variable not set correctly"
  echo "Got: '${env}'"
  exit 1
fi
"#
    )
}
```

Will move it to dev-deps though :+1:

not-matthias:
Added `rstest_reuse` and moved the constants with the test scripts/env vars into the array. 

Note: the last 2 commits should be ignored, as they are merged from https://github.com/CodSpeedHQ/runner/pull/99 (not sure why Github didn't recognize that, I'll close both once this PR is merged)

art049 (approved):
Nice! @GuillaumeLagrange I let you do the final review

GuillaumeLagrange (approved):
lgtm to me!

PR #106, opened 2025-08-28
fix(go): prevent running with valgrind

not-matthias:
Works now as expected:
```
> codspeed run --skip-upload -- go test -bench=.                                             nix-shell-env 
Unofficially supported system: nixos 25.05. Continuing with best effort support.

►►► Preparing the environment 
Valgrind is already installed with the correct version, skipping installation
Environment ready

►►► Running the benchmarks 
WARNING: Go benchmarks can only be run with walltime
Error: failed to execute the benchmark process, exit code: 1
```


src/run/runner/helpers/introspected_golang/go.sh
```diff
@@ -11,6 +11,13 @@ debug_log() {
 debug_log "Called with arguments: $*"
 debug_log "Number of arguments: $#"
 
+
+# Currently only walltime is supported
+if [ "${CODSPEED_RUNNER_MODE:-}" != "walltime" ]; then
+    echo "WARNING: Go benchmarks can only be run with walltime"
```
art049: It's an error since we're exiting and we can tweak a bit the wording
```suggestion
    echo "CRITICAL: Go benchmarks can only be run with the walltime instrument"
```

art049 (approved):

PR #107, opened 2025-08-28
fix: compute proper load bias

not-matthias:


not-matthias:
> Nice, lmk when you have additional tests for Python + Rust, as we discussed

Verified the changes for Cpp, Rust and Golang. I've added test cases for a failing example and made sure the flamegraphs are the same.

The python runner implementation isn't fully completely (see #103 ) + my uv env is completely broken so I skipped it for now. But I'm very confident that it will work for Python as well.

PR #108, opened 2025-08-31
Use introspected ndoe in walltime mode

GuillaumeLagrange:


src/run/runner/helpers/introspected_golang/mod.rs
```diff
@@ -5,7 +5,7 @@ const INTROSPECTED_GO_SCRIPT: &str = include_str!("go.sh");
 
 /// Creates the `go` script that will replace the `go` binary while running
 /// Returns the path to the script folder, which should be added to the PATH environment variable
```
not-matthias, line 7: Can you also update this comment since it's identical with the one on `setup_introspected_nodejs` that you changed?

not-matthias (approved):
LGTM!

src/run/runner/helpers/introspected_golang/mod.rs
GuillaumeLagrange: I had changed it by mistake, I reverted it

PR #110, opened 2025-09-01
Make the runner mode argument mandatory and enable perf by default

GuillaumeLagrange:


src/run/mod.rs
```diff
@@ -52,7 +52,7 @@ pub enum UnwindingMode {
 pub struct PerfRunArgs {
     /// Enable the performance runner, which uses `perf` to collect performance data.
```
art049:
```suggestion
    /// Enable the linux perf profiler to collect granular performance data.
```

src/run/mod.rs
```diff
@@ -52,7 +52,7 @@ pub enum UnwindingMode {
...
     /// This is only supported on Linux.
-    #[arg(long, env = "CODSPEED_PERF_ENABLED", default_value_t = false)]
+    #[arg(long, env = "CODSPEED_PERF_ENABLED", default_value_t = true)]
     enable_perf: bool,
```
art049, line 56: Does it work if perf is not available? Just to make sure this wont break any workflow if people don't have perf or an incorrect version?

art049: cc @not-matthias

not-matthias: We have an assert in the upload part where we check that `bench_count == perf_data_count` which could fail.
https://github.com/CodSpeedHQ/runner/blob/6e44b77cef32bf89d1b3b92a2c86d7747643e2d3/src/run/runner/wall_time/perf/mod.rs#L188-L193

Also, if someone uses an old integration which doesn't use the FIFO yet then we'll crash due to not receiving integration metadata: https://github.com/CodSpeedHQ/runner/blob/6e44b77cef32bf89d1b3b92a2c86d7747643e2d3/src/run/runner/wall_time/perf/mod.rs#L387-L390

art049: Maybe we can change the assert `bench_count == perf_data_count` to a warning.

> Also, if someone uses an old integration which doesn't use the FIFO yet then we'll crash due to not receiving integration metadata.

Can we find a way around this one though?

not-matthias: In theory, we could change the integration to `Option<(String, String)>` or create placeholder values since we won't create any perf.data files if the integration doesn't support it. Changing it to `Option` would also require a change on the backend.
EDIT: Or just skip creating the perf metadata file in that case.

art049: Is the integration still needed here? We used it for valgrind IIRC but now we also have it in the results.json I think

art049: > Or just skip creating the perf metadata file in that case.

Yes, it would definitely be simpler 👍

not-matthias: > Is the integration still needed here? We used it for valgrind IIRC but now we also have it in the results.json I think

Oh true, since we're parsing the results.json in https://github.com/CodSpeedHQ/codspeed/pull/1372, we could get rid of the integration cmd entirely.

art049: It might be slightly more complicated for the valgrind side thant purely skipping the metadata creation as you mentioned but we can mark it as deprecated already so we stop using it

GuillaumeLagrange: I've adressed this in the latest push 
https://github.com/CodSpeedHQ/runner/pull/110/commits/1e6c4de0b1e9107b1d48056da8f7fb0955857b1c#diff-e34c2f603add48260c577dfad1d936056a33f4a88ae0396b62e76cc2643f18f5R209-R389

Is this OKFY ?

CLAUDE.md
art049, line 1: let's do it with AGENTS.md instead

src/run/mod.rs
art049: yes

PR #114, opened 2025-09-15
fix: find all installed python versions

not-matthias:


not-matthias:
Closing this since it was only used for testing.

PR #115, opened 2025-09-16
fix: forward go runner exit status

not-matthias:


src/run/runner/helpers/introspected_golang/go.sh
```diff
@@ -47,8 +47,10 @@ if [ "$1" = "test" ]; then
     debug_log "Full command: RUST_LOG=info $GO_RUNNER $*"
 
     "$GO_RUNNER" "$@"
+    exit $?
 else
     debug_log "Detected non-test command ('$1'), routing to standard go binary"
     debug_log "Full command: $REAL_GO $*"
     "$REAL_GO" "$@"
+    exit $?
```
art049: nitpick but maybe we could just have one exit after if else fi

PR #116, opened 2025-09-16
fix: allow perf to be started/resumed

not-matthias:
- [ ] extensively test for all languages that we don't add regresssions here


not-matthias:
Not relevant anymore, closing this.

PR #117, opened 2025-09-16
COD-1345: Do not gzip walltime reports

adriencaccia:


src/run/run_environment/local/provider.rs
```diff
@@ -153,7 +153,7 @@ impl RunEnvironmentProvider for LocalProvider {
         let run_environment_metadata = self.get_run_environment_metadata()?;
 
         Ok(UploadMetadata {
-            version: Some(6),
+            version: Some(7),
```
art049: Let's store this in a const

src/run/run_environment/provider.rs
```diff
@@ -66,19 +66,20 @@ pub trait RunEnvironmentProvider {
...
 
         Ok(UploadMetadata {
-            version: Some(6),
+            version: Some(7),
             tokenless: config.token.is_none(),
             repository_provider: self.get_repository_provider(),
             run_environment_metadata,
-            profile_md5: archive_hash.into(),
+            profile_md5: profile_archive.hash().into(),
+            content_encoding: profile_archive.content_encoding(),
```
art049: for consistency with profile_md5 maybe this should be `profile_encoding`?

src/request_client.rs
```diff
@@ -16,4 +16,10 @@ lazy_static! {
         ExponentialBackoff::builder().build_with_max_retries(UPLOAD_RETRY_COUNT)
     ))
     .build();
+
+    // Client without retry middleware for streaming uploads (can't be cloned)
+    pub static ref STREAMING_CLIENT: reqwest::Client = ClientBuilder::new()
+        .user_agent("codspeed-runner")
+        .build()
+        .unwrap();
```
art049, line 24: would be great to find a way to have the retries even when streaming the body

art049 (changes requested):

src/request_client.rs
adriencaccia: As seen in https://github.com/TrueLayer/reqwest-middleware/pull/51/files?w=1#diff-474abab8d90fcf4a3811b4faa74eff00faa8c9c190c9fd2e47991de8b2f4d64eR38-R49, it is not possible out of the box.
We will need to create a middleware manually that implements streaming. I will take a look

PR #118, opened 2025-09-16
fix: only enable debug logs GH action is debugged

not-matthias:


src/run/run_environment/github_actions/logger.rs
```diff
@@ -4,12 +4,30 @@ use crate::{
...
-pub struct GithubActionLogger;
+pub struct GithubActionLogger {
+    log_level: LevelFilter,
+}
+
+impl GithubActionLogger {
+    pub fn new() -> Self {
+        // Only enable debug logging if it's enabled in GitHub Actions.
+        let log_level = if env::var("RUNNER_DEBUG").unwrap_or_default() == "1" {
+            LevelFilter::Debug
```
art049: this should be set to trace since this was the default before

src/run/run_environment/github_actions/logger.rs
```diff
@@ -4,12 +4,30 @@ use crate::{
...
 /// See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions
-pub struct GithubActionLogger;
+pub struct GithubActionLogger {
+    log_level: LevelFilter,
+}
+
+impl GithubActionLogger {
+    pub fn new() -> Self {
+        // Only enable debug logging if it's enabled in GitHub Actions.
+        let log_level = if env::var("RUNNER_DEBUG").unwrap_or_default() == "1" {
```
art049, line 20: Do you have docs on this RUNNER_DEBUG flag?

not-matthias: It's mentioned in the docs here: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context, i'll add a link in the code

art049: Ah thanks. Actually this page is about the env variables: https://docs.github.com/en/actions/reference/workflows-and-actions/variables

art049 (approved):

PR #119, opened 2025-09-19
Cod 1364 runner sample without splitting perfdata

not-matthias:
Changes in this PR: 
- Create `runner-shared` crate which is used in the `perf-parser` crate
- Add support for the perf v2 format
- Add `SetVersion` command to detect incompatible integrations

### DIFF
--- a/crates/runner-shared/src/fifo.rs
+++ b/crates/runner-shared/src/fifo.rs
@@ -1,0 +1,33 @@
+//! WARNING: Has to be in sync with `instrument-hooks`.
+
+pub const RUNNER_CTL_FIFO: &str = "/tmp/runner.ctl.fifo";
+pub const RUNNER_ACK_FIFO: &str = "/tmp/runner.ack.fifo";
+
+pub const CURRENT_PROTOCOL_VERSION: u64 = 1;
+
+/// The different markers that can be set in the perf.data.
+///
+/// `SampleStart/End`: Marks the start and end of a sampling period. This is used to differentiate between benchmarks.
+/// `BenchmarkStart/End`: Marks the start and end of a benchmark. This is used to measure the duration of a benchmark, without the benchmark harness code.
+#[derive(
+    serde::Serialize, serde::Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Copy, Clone,
+)]
+pub enum MarkerType {
+    SampleStart(u64),
+    SampleEnd(u64),
+    BenchmarkStart(u64),
+    BenchmarkEnd(u64),
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Debug, PartialEq)]
+pub enum Command {
+    CurrentBenchmark { pid: i32, uri: String },
+    StartBenchmark,
+    StopBenchmark,
+    Ack,
+    PingPerf,
+    SetIntegration { name: String, version: String },
+    Err,
+    AddMarker { pid: i32, marker: MarkerType },
+    SetVersion(u64),
+}
--- a/src/run/runner/wall_time/perf/mod.rs
+++ b/src/run/runner/wall_time/perf/mod.rs
@@ -9,35 +9,34 @@ use crate::run::runner::helpers::setup::run_with_sudo;
 use crate::run::runner::valgrind::helpers::ignored_objects_path::get_objects_path_to_ignore;
 use crate::run::runner::valgrind::helpers::perf_maps::harvest_perf_maps_for_pids;
 use crate::run::runner::wall_time::perf::jit_dump::harvest_perf_jit_for_pids;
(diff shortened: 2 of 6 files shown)


crates/runner-shared/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+// This will be accessed from the `parse_callgraph` lambda.
```
art049: This shouldn't appear here

```suggestion
```

crates/runner-shared/src/fifo.rs
```diff
@@ -0,0 +1,33 @@
+//! WARNING: Has to be in sync with `codspeed-rust/codspeed`.
```
art049: since this is shared, codspeed-rust could use this directly no?

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -126,10 +116,10 @@ impl PerfRunner {
             "sh",
             "-c",
             &format!(
-                "perf record {quiet_flag} --user-callchains --freq=999 --switch-output --control=fifo:{},{} --delay=-1 -g --call-graph={cg_mode} --output={} -- {bench_cmd}",
+                "perf record {quiet_flag} --timestamp -k CLOCK_MONOTONIC --user-callchains --freq=999 --control=fifo:{},{} --delay=-1 -g --call-graph={cg_mode} --output={} -- {bench_cmd}",
                 perf_fifo.ctl_fifo_path.to_string_lossy(),
                 perf_fifo.ack_fifo_path.to_string_lossy(),
-                perf_file.path().to_string_lossy(),
+               PERF_DATA_PATH
```
art049: The argument list can have improved readability, similarly to what we have in valgrind measure.rs.

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -230,8 +181,8 @@ impl PerfRunner {
     #[cfg(target_os = "linux")]
     fn process_memory_mappings(
         pid: u32,
-        symbols_by_pid: &mut HashMap<u32, ProcessSymbols>,
-        unwind_data_by_pid: &mut HashMap<u32, Vec<UnwindData>>,
+        symbols_by_pid: &mut HashMap<i32, ProcessSymbols>,
+        unwind_data_by_pid: &mut HashMap<i32, Vec<UnwindData>>,
```
art049: is this change intended? we lack of consistency between the pid types defined here.
I think you could create a type alias `pid_t` that maps to the type we chose similarly to what's available in c stdlib

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -269,8 +220,9 @@ impl PerfRunner {
                 continue;
             }
 
+            let pid_i32 = pid as i32;
```
art049: The previous issue leads to this useless cast

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -378,16 +332,33 @@ impl PerfRunner {
...
+                }
+                FifoCommand::SetVersion(protocol_version) => {
+                    if protocol_version != runner_shared::fifo::CURRENT_PROTOCOL_VERSION {
+                        panic!(
+                            "Integration is outdated, please update the codspeed integration to the latest version."
+                        );
+                    }
+
+                    runner_fifo.send_cmd(FifoCommand::Ack).await?;
+                }
```
art049: I think there are 2 cases to distinguish here: received > current and received < current
Probably the error message should differ

art049 (changes requested):

src/run/runner/wall_time/perf/mod.rs
not-matthias: I changed it because I realized that we're using i32 in the backend, valgrind and various helper functions (e.g. `harvest_perf_maps`) which makes conversions more cumbersome. I'll update the pid field as well :+1:

crates/runner-shared/src/fifo.rs
not-matthias: we're actually now only using this in instrument-hooks. i'll update the comment 👍

art049:
> > It's weird you had to fix the clippy issues since we have this in the recommit hooks:
> > ```
> >       - id: clippy
> >         args: [--all-targets, --, -D, warnings]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Did the rust version or anything else change?
> 
> Ahh, I think I'm using a newer Rust version. Should I revert the commit?

Anyway we'd have to do those changes at some point. Are you using rustup? Normally it picks up the rust-toolchain.toml automatically

PR #120, opened 2025-09-22
fix: add base addr to symbol offset

not-matthias:
Changes in this PR: 
- Merge the load bias and base address calculation of `unwind_info` and `perf_map` 
  - (we previously had 2 separate implementations - it's now easier to verify and maintain)
- Add the base address to the symbol offset, rather than the load bias (only if it's not already an absolute address)
- Add test to for `the_algorithms` to ensure we get the right addresses


crates/runner-shared/src/unwind_data.rs
art049, line 1: It would be simpler overall to implement `From<UnwindDataV1> `on `UnwindDataV2` so in the future, we won't have to think about the other versions but just about migrating from the current one to the new one.

crates/runner-shared/src/unwind_data.rs
```diff
@@ -5,45 +5,153 @@ use core::{
...
+    pub fn eh_frame_svma(&self) -> &Range<u64> {
+        match self {
+            UnwindData::V1(data) => &data.eh_frame_svma,
+            UnwindData::V2(data) => &data.eh_frame_svma,
+        }
+    }
+
+    pub fn save_to<P: AsRef<std::path::Path>>(&self, folder: P, pid: i32) -> anyhow::Result<()> {
+        let unwind_data_path = folder.as_ref().join(format!(
+            "{}_{:x}_{:x}.unwind2",
```
art049: `unwind.bin` would be a bit nicer

src/run/runner/wall_time/perf/elf_helper.rs
```diff
@@ -0,0 +1,202 @@
...
+///  - mach-O libraries in the dyld shared cache have a __TEXT segment at some
+///    non-zero address in the cache.
+///  - ELF executables can have non-zero base addresses, e.g. 0x200000 or 0x400000.
+///  - Kernel ELF binaries ("vmlinux") have a large base address such as
+///    0xffffffff81000000. Moreover, the base address seems to coincide with the
+///    vmaddr of the .text section, which is readily-available in perf.data files
+///    (in a synthetic mapping called "[kernel.kallsyms]_text").
+///
+/// Credits: https://github.com/mstange/samply/blob/4a5afec57b7c68b37ecde12b5a258de523e89463/samply-symbols/src/shared.rs#L513-L566
+pub fn relative_address_base(object_file: &object::File) -> u64 {
```
GuillaumeLagrange, line 165: this can be private, to better scope the complexity

crates/runner-shared/src/unwind_data.rs
```diff
@@ -5,45 +5,153 @@ use core::{
...
+    pub base_avma: u64,
+    pub eh_frame_hdr: Vec<u8>,
+    pub eh_frame_hdr_svma: Range<u64>,
+
+    pub eh_frame: Vec<u8>,
+    pub eh_frame_svma: Range<u64>,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct UnwindDataV2 {
```
GuillaumeLagrange, line 106: Could we add a timestamp to this one ? Could be an option, where `None` would mean "from the start of the program"

In the impl From V1 it would translate to None

crates/runner-shared/src/unwind_data.rs
GuillaumeLagrange: Ideally, versioning of this would be completely local to a deserialization layer, and applicative code would only manipulate the latest version without being able to import the old version.

crates/runner-shared/src/unwind_data.rs
not-matthias: I called it `unwind2` since it's unwind v2 . if we call it `unwind.bin` it's going to be a bit more tricky since the extension will then be `bin` which is less clear and could conflict.

crates/runner-shared/src/unwind_data.rs
not-matthias: I had a version where I re-exported `UnwindDataV2` (or whatever is the latest version) as `UnwindData` which can then be used in the runner. 

But there was an issue where we would need to `.clone()` it when converting it into `UnwindDataCompat`. Maybe we can have `UnwindDataCompat` use lifetimes with borrowed values. Then we can always only use the UnwindDataV2 and hide the de/serialization.

crates/runner-shared/src/unwind_data.rs
not-matthias: I replaced `UnwindDataV1` with `UnwindDataV2`. But I'll add a new V2 for the timestamp since that would break all of the snapshots currently in the monorepo. But this is good, since we can then ensure the compatibility layer works.

src/run/runner/wall_time/perf/elf_helper.rs
not-matthias: we're actually using this in `unwind_data.rs`

art049 (approved):

PR #128, opened 2025-10-13
feat: save file/line info for symbols

not-matthias:


.gitattributes
art049, line 1: maybe we could put all the snaps on lfs?

src/run/runner/wall_time/perf/debug_info.rs
```diff
@@ -0,0 +1,226 @@
...
+        std::fs::write(debug_info_path, json)?;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::run::runner::wall_time::perf::perf_map::ModuleSymbols;
```
GuillaumeLagrange: Aren't we already importing this in the super module ?

src/run/runner/wall_time/perf/debug_info.rs
```diff
@@ -0,0 +1,206 @@
...
+
+        Self { pid, modules }
+    }
+
+    pub fn save_to<P: AsRef<std::path::Path>>(&self, folder: P) -> anyhow::Result<()> {
+        if self.modules.is_empty() {
+            return Ok(());
+        }
+
+        let debug_info_path = folder.as_ref().join(format!("{}.debuginfo", self.pid));
```
GuillaumeLagrange: As discussed, we should put this into the perf.metadata or something, in order to avoid creating another file format that we need to maintain.

GuillaumeLagrange (changes requested):
Will finish the review when the file format stuff is sorted out, but OLGTM, minus the fixup not fixed up

PR #140, opened 2025-10-23
fix(sudo): handle root user case

fargito:


art049:
this still need to be fixed

PR #142, opened 2025-10-24
chore: use info instead of warn for some cache and valgrind setup logs

adriencaccia:


src/run/runner/valgrind/setup.rs
art049, line 1: for those, a warning seems to make sense

PR #145, opened 2025-11-03
feat: add --config-name argument to allow multiple configs

GuillaumeLagrange:
Allows using the codspeed CLI in different environments without having to re-log in everytime

Helpful bindings to go along

```
        eval $(op signin)
        export CODSPEED_CONFIG_NAME=dev
        export CODSPEED_API_URL=$(op read "op://Private/codspeed_urls/dev_api_url")
        export CODSPEED_UPLOAD_URL=$(op read "op://Private/codspeed_urls/dev_upload_url")
```

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -40,6 +40,12 @@ pub struct Cli {
     #[arg(long, env = "CODSPEED_OAUTH_TOKEN", global = true, hide = true)]
     pub oauth_token: Option<String>,
 
+    /// The configuration name to use
+    /// If provided, the configuration will be loaded from ~/.config/codspeed/{config-name}.yaml
+    /// Otherwise, loads from ~/.config/codspeed/config.yaml
+    #[arg(long, env = "CODSPEED_CONFIG_NAME", global = true)]
+    pub config_name: Option<String>,
+
     /// The directory to use for caching installed tools
     /// The runner will restore cached tools from this directory before installing them.
     /// After successful installation, the runner will cache the installed tools to this directory.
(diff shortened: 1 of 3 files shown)


src/app.rs
```diff
@@ -83,7 +90,7 @@ pub async fn run() -> Result<()> {
         Commands::Run(args) => {
             run::run(args, &api_client, &codspeed_config, setup_cache_dir).await?
         }
-        Commands::Auth(args) => auth::run(args, &api_client).await?,
+        Commands::Auth(args) => auth::run(args, &api_client, cli.config_name.as_deref()).await?,
```
not-matthias, line 93: I think it might make more sense to pass `codspeed_config` here since we already instantiate it. The only downside is that we'd have to also add the config_name in the struct to ensure we can call `persist()`. 

(lmk what you think, we can also do it like you did)

src/app.rs
```diff
@@ -40,6 +40,12 @@ pub struct Cli {
     #[arg(long, env = "CODSPEED_OAUTH_TOKEN", global = true, hide = true)]
     pub oauth_token: Option<String>,
 
+    /// The configuration name to use
+    /// If provided, the configuration will be loaded from ~/.config/codspeed/{config-name}.yaml
+    /// Otherwise, loads from ~/.config/codspeed/config.yaml
+    #[arg(long, env = "CODSPEED_CONFIG_NAME", global = true)]
+    pub config_name: Option<String>,
```
not-matthias, line 47: i think `config_profile` might be more suitable since other programs have the same naming (rust, assume, ...)

GuillaumeLagrange: as discussed, let's keep `config_name` since `profile` is a business term for us and we don't want to pollute it with something minor like this

src/app.rs
GuillaumeLagrange: for now I'm sticking with this because I don't want to bother changing the yaml serialization/deserialization by including the config name in the struct.

It's simpler to keep the value from clap as the source of truth when loading/saving the config to disk.

In a future major version, the best way to do all this would be to have the config on disk be a single file, and nest the information inside top level yaml keys that are the profile/config names

PR #148, opened 2025-11-07
fix: valgrind cache restore fails

not-matthias:


src/run/runner/helpers/apt.rs
```diff
@@ -225,3 +233,29 @@ fn save_to_cache(system_info: &SystemInfo, cache_dir: &Path, packages: &[&str])
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs;
+    use tempfile::TempDir;
+
+    #[test]
+    fn test_copy_files_recursively() {
```
GuillaumeLagrange: not of fan of unit tests that require sudo permissions, imo here we should not have these unit tests here

not-matthias: I can also just modify the cmd and create a ticket to add unit tests for the caching in the runner, wdyt?

adriencaccia (approved):

src/run/runner/wall_time/perf/fifo.rs
```diff
@@ -32,6 +32,10 @@ fn get_pipe_open_options() -> TokioPipeOpenOptions {
 
 impl RunnerFifo {
     pub fn new() -> anyhow::Result<Self> {
+        // Delete any existing FIFOs to avoid hangs
```
GuillaumeLagrange: I dont get this, didnt we say yesterday that it was no longer necessary ?

not-matthias: the commit is already gone :P created the branch on top of the wrong branch

src/run/runner/helpers/apt.rs
adriencaccia: For some reason this test fails when I try to run it locally on my mac

```
    failures:
        run::runner::helpers::apt::tests::test_copy_files_recursively

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 132 filtered out; finished in 0.02s
    
  stderr ───

    thread 'run::runner::helpers::apt::tests::test_copy_files_recursively' panicked at src/run/runner/helpers/apt.rs:250:92:
    called `Result::unwrap()` on an `Err` value: Failed to execute command with sudo: "sudo" "--non-interactive" "bash" "-c" "cp -r /var/folders/dh/mxjcx6nn6_5_5dv3zlxqhvgc0000gn/T/.tmpjoQmVj/* /var/folders/dh/mxjcx6nn6_5_5dv3zlxqhvgc0000gn/T/.tmp9VikXe"
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Might be best to simply remove it, as as @GuillaumeLagrange pointed out, it's not great to rely on `sudo` in tests

GuillaumeLagrange: Yes let's do this

It would be some sort of integration test in the runner, but the more we can avoid sudo-requiring unit test the better.

Think about the user experience of someone coming to this repo and wanting to contribute something, running `cargo test` and being prompted for a sudo password :grimacing:

adriencaccia (approved):

PR #151, opened 2025-11-10
fix: use GITHUB_WORKSPACE env var when computing root path

GuillaumeLagrange:
Related to https://github.com/CodSpeedHQ/action/issues/153

### DIFF
--- a/src/run/run_environment/github_actions/provider.rs
+++ b/src/run/run_environment/github_actions/provider.rs
@@ -1,3 +1,4 @@
+use git2::Repository;
 use lazy_static::lazy_static;
 use regex::Regex;
 use serde_json::Value;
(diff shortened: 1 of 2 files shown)


src/run/run_environment/github_actions/provider.rs
```diff
@@ -209,6 +218,24 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
             metadata,
         })
     }
+
+    fn get_commit_hash(&self, repository_root_path: &str) -> Result<String> {
+        let repo = Repository::open(repository_root_path).context(format!(
+            "Failed to open repository at path: {repository_root_path}\n\
+            Make sure git is installed, and that `actions/checkout` used git to fetching the repository\n\
```
not-matthias: small typo:
```suggestion
            Make sure git is installed, and that `actions/checkout` used git to fetch the repository\n\
```

src/run/run_environment/github_actions/provider.rs
```diff
@@ -209,6 +218,24 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
...
+            If necessary, install git before running `actions/checkout`.\n\
+            If running in docker and running into permission issues, you may need to also run \
```
not-matthias:
```suggestion
            If you run into permission issues when running in Docker, you may need to also run \
```

src/run/run_environment/provider.rs
```diff
@@ -94,16 +80,19 @@ pub trait RunEnvironmentProvider {
             run_part: self.get_run_provider_run_part(),
         })
     }
-}
 
-#[cfg(test)]
-mod tests {
-    use super::*;
+    /// Returns the HEAD commit hash of the repository at the given path.
+    fn get_commit_hash(&self, repository_root_path: &str) -> Result<String> {
```
not-matthias, line 85: Is there any reason you removed the `test_get_commit_hash` test? 
```rust
    #[test]
    fn test_get_commit_hash() {
        let commit_hash = get_commit_hash(env!("CARGO_MANIFEST_DIR")).unwrap();
        // ensure that the commit hash is correct, thus it has 40 characters
        assert_eq!(commit_hash.len(), 40);
    }
```

GuillaumeLagrange: oops, restored the test

PR #152, opened 2025-11-13
feat: accept simulation as runner mode

GuillaumeLagrange:


src/run/mod.rs
```diff
@@ -138,6 +138,7 @@ pub struct RunArgs {
 #[serde(rename_all = "lowercase")]
 pub enum RunnerMode {
     Instrumentation,
```
adriencaccia, line 141: It would be a good use for [`#[deprecated]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute)

src/run/runner/helpers/env.rs
```diff
@@ -6,6 +6,14 @@ pub fn get_base_injected_env(
...
     profile_folder: &Path,
 ) -> HashMap<&'static str, String> {
+    let runner_mode_internal_env_value = match mode {
+        RunnerMode::Instrumentation => "instrumentation",
+        // While the runner now deprecates the usage of instrumentation with a message, we
+        // internally still use instrumentation temporarily to give time to users to upgrade their
+        // integrations to a version that accepts both instrumentation and simulation.
+        RunnerMode::Simulation => "instrumentation",
+        RunnerMode::Walltime => "walltime",
+    };
```
adriencaccia, line 18: Let's have a TODO comment to remember to remove this logic in the next major version

PR #154, opened 2025-11-13
Enable perf event compression and add more events to the monitoring

GuillaumeLagrange:
- Don't merge before we add zstd compression to perf sampling

### DIFF
--- a/src/executor/wall_time/perf/perf_executable.rs
+++ b/src/executor/wall_time/perf/perf_executable.rs
@@ -1,3 +1,5 @@
+use runner_shared::perf_event::PerfEvent;
+
 use crate::prelude::*;
 use std::path::Path;
 
@@ -67,15 +69,7 @@ pub fn get_working_perf_executable() -> Option<OsString> {
 /// Detects if the required perf events are available on this system.
 /// Returns the flags to pass to perf record command if they are available, otherwise returns None.
 pub fn get_event_flags(perf_executable: &OsString) -> anyhow::Result<Option<String>> {
-    const CYCLES_EVENT_NAME: &str = "cycles";
-    const CACHE_REFERENCES_EVENT_NAME: &str = "cache-references";
-    const CACHE_MISSES_EVENT_NAME: &str = "cache-misses";
-
-    let perf_events = [
-        CYCLES_EVENT_NAME,
-        CACHE_REFERENCES_EVENT_NAME,
-        CACHE_MISSES_EVENT_NAME,
-    ];
+    let perf_events = PerfEvent::all_events();
 
     let output = Command::new(perf_executable)
         .arg("list")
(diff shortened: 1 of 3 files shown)


src/run/runner/wall_time/perf/perf_executable.rs
```diff
@@ -62,3 +62,39 @@ pub fn get_working_perf_executable() -> Option<OsString> {
...
+    const CACHE_MISSES_EVENT_NAME: &str = "cache-misses";
+    let output = Command::new(perf_executable)
+        .arg("list")
+        .output()
+        .context("Failed to run perf list")?;
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+
+    // Check for the three required events
+    let has_cycles = stdout.lines().any(|line| line.contains(CYCLES_EVENT_NAME));
```
not-matthias: This check can easily return true, even if we don't support cycles because it's only a substring match. For example: 
```
  ref-cycles                                         [Hardware event]
```
would also return `true` despite not being the same event (not sure if it's that likely or if there are any systems where it's not available).

Maybe do it like this:
```suggestion
    let has_cycles = stdout.lines().any(|line| line.trim().starts_with(CYCLES_EVENT_NAME));
```

not-matthias: nit: this could also be a list of `&[&str]` which we could iterate over, then we'd simplify the logic since it's the same check.

GuillaumeLagrange: Much better indeed!

src/executor/wall_time/perf/perf_executable.rs
```diff
@@ -62,3 +65,78 @@ pub fn get_working_perf_executable() -> Option<OsString> {
...
+            !stdout.lines().any(|line| {
+                line.split_whitespace()
+                    .any(|word| word == event.to_perf_string())
+            })
+        })
+        .copied()
+        .collect();
+
+    if !missing_events.is_empty() {
+        debug!(
```
art049: should be a warning

src/executor/wall_time/perf/perf_executable.rs
```diff
@@ -62,3 +65,78 @@ pub fn get_working_perf_executable() -> Option<OsString> {
...
+    let has_zstd = stdout
+        .lines()
+        .any(|line| line.to_lowercase().contains("zstd: [ on"));
+
+    if has_zstd {
+        debug!("perf supports zstd compression");
+        // 3 is a widely adopted default level (AWS Athena, Python, ...)
+        Ok(Some("--compression-level=3".to_string()))
+    } else {
+        debug!("perf does not support zstd compression");
```
art049: warning as well would be nice

art049 (approved):

PR #156, opened 2025-11-14
COD-1469: Remove the CodSpeed token from the setup

fargito:


src/run/run_environment/github_actions/provider.rs
```diff
@@ -236,6 +245,46 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
...
+        let request_url = get_env_variable("ACTIONS_ID_TOKEN_REQUEST_URL").ok();
+
+        if request_token.is_none() || request_url.is_none() {
+            warn!(
+                "The Workflow is missing the 'id-token: write' permission required to request an OIDC token. \
+                Please refer to https://docs.github.com/en/actions/how-to-guides/security-hardening-your-workflows/about-security-hardening-with-openid-connect for more information."
+            );
+
+            return None;
+        }
```
adriencaccia, line 319: - [x] The linked docs page does not exist
- [x] Change the waning message to a link to our docs, and explain the benefit for the users, that they can remove `CODSPEED_TOKEN` in favor of `id-token: write`

src/run/mod.rs
```diff
@@ -195,6 +195,13 @@ pub async fn run(
...
         debug!("Using the token from the CodSpeed configuration file");
         config.set_token(codspeed_config.auth.token.clone());
+    } else {
+        let oidc_token = provider.get_oidc_token().await;
+        if oidc_token.is_some() {
+            debug!("Using OIDC token for authentication");
+
+            config.set_token(oidc_token);
+        }
     }
```
adriencaccia, line 214: Let's not overwrite the token is it was explicitly set

src/run/run_environment/github_actions/provider.rs
```diff
@@ -236,6 +245,46 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
...
+
+        let request_url = request_url.unwrap();
+        let request_url = format!("{request_url}&audience=codspeed");
```
adriencaccia: Put the audience in a constant and change the value to `api://codspeed`

adriencaccia (changes requested):

src/run/run_environment/github_actions/provider.rs
```diff
@@ -236,6 +245,46 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
...
+        let request_token = request_token.unwrap();
+
+        Client::new()
```
adriencaccia: Let's add a similar retrying logic as they do in the OIDC client in https://github.com/actions/toolkit/blob/dcae869a030cef8949f37078ccc240752fb69338/packages/core/src/oidc-utils.ts#L14

src/run/run_environment/github_actions/provider.rs
fargito: https://github.com/CodSpeedHQ/docs/pull/33

src/run/run_environment/gitlab_ci/provider.rs
```diff
@@ -175,6 +177,13 @@ impl RunEnvironmentProvider for GitLabCIProvider {
             metadata: BTreeMap::new(),
         })
     }
+
+    /// For GitLab CI, OIDC tokens must be passed via env variable.
+    ///
+    /// See:
+    /// - https://docs.gitlab.com/integration/openid_connect_provider/
+    /// - https://docs.gitlab.com/ci/secrets/id_token_authentication/
+    async fn set_oidc_token(&self, _config: &mut Config) {}
```
adriencaccia: It is unclear that it will work because in the docs we recommend to use `CODSPEED_TOKEN` as the value of the token, so `config.token` will contain the correct token.
Add some explanation here

src/run/run_environment/github_actions/provider.rs
```diff
@@ -236,6 +249,80 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
...
+            // Only display a warning if the run is not from a fork
+            if !self.is_head_repo_fork {
+                warn!(
+                    "CodSpeed was unable to request an OIDC token for this job.\n
+                    Add the `id-token: write` permission to your workflow, and benefit from enhanced security and faster processing times.\n
+                    Learn more at https://codspeed.io/docs/integrations/ci/github-actions#openid-connect-oidc-authentication"
+                );
+            }
+
+            return;
```
adriencaccia: Since tokenless works for public repositories, we this should only happen for private repositories.
Furthermore, since uploading on a private repository requires a token, if we do not have a token by this point, the subsequent upload will fail.
Thus let's simply bail here

fargito: I don't think we want to do that for now. Since it is not possible to know in the runner if the repository is private or public, we could end up in this branch with a public repo that currently uses a tokenless setup and does not have `id-token: write` permissions set. This is why I displayed it with a warning.

fargito: I'll add a comment to make that clearer

adriencaccia: Of course my comment only makes sense if we can know if a repository is private or public.
And we do have the information available in the file stored at `GITHUB_EVENT_PATH`, in the `repository.private` boolean field

fargito: Oh right, we don't use it yet. That will be a lot better indeed

src/run/run_environment/github_actions/provider.rs
```diff
@@ -53,13 +67,20 @@ impl TryFrom<&Config> for GitHubActionsProvider {
...
         let is_pr = PR_REF_REGEX.is_match(&ref_);
-        let head_ref = if is_pr {
-            let github_event_path = get_env_variable("GITHUB_EVENT_PATH")?;
-            let github_event = fs::read_to_string(github_event_path)?;
-            let github_event: Value = serde_json::from_str(&github_event)
-                .expect("GITHUB_EVENT_PATH file could not be read");
+
+        let is_repository_private = github_event["repository"]["private"]
+            .as_bool()
+            .unwrap_or(true);
```
adriencaccia: This is arbitrary. I would prefer to `unwrap` here as we did for the PR event fields

src/run/run_environment/github_actions/provider.rs
```diff
@@ -236,6 +260,93 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
...
+    /// - https://docs.github.com/en/actions/concepts/security/openid-connect
+    /// - https://docs.github.com/en/actions/reference/security/oidc#methods-for-requesting-the-oidc-token
+    async fn set_oidc_token(&self, config: &mut Config) -> Result<()> {
+        // Check if a static token is already set
+        if config.token.is_some() {
+            info!(
+                "CodSpeed now supports OIDC tokens for authentication.\n
+                Add the `id-token: write` permission to your workflow, remove the token from your configuration, and benefit from enhanced security.\n
+                Learn more at https://codspeed.io/docs/integrations/ci/github-actions#openid-connect-oidc-authentication"
+            );
```
adriencaccia, line 288: Let's change the text by starting with the benefits of OIDC in comparison to the static token.

src/run/run_environment/github_actions/provider.rs
```diff
@@ -236,6 +260,93 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
...
+                bail!(
+                    "Unable to retrieve OIDC token for authentication. \n
+                    Make sure your workflow has the `id-token: write` permission set. \n
+                    See https://codspeed.io/docs/integrations/ci/github-actions#openid-connect-oidc-authentication"
+                )
+            }
+
+            info!(
+                "CodSpeed now supports OIDC tokens for authentication.\n
+                Add the `id-token: write` permission to your workflow, and benefit from enhanced security and faster processing times.\n
```
adriencaccia: Same thing for this text here, it should contain the benefits first.

PR #160, opened 2025-11-20
feat: add --allow-empty run option

GuillaumeLagrange:
Mostly aimed for action runner CI which just tests cli flags it passes and does not run benches

### DIFF
--- a/src/run/runner/wall_time/helpers.rs
+++ b/src/run/runner/wall_time/helpers.rs
@@ -11,10 +11,15 @@ fn add_empty_result_error_explanation(error_details: &str) -> String {
 }
 
 /// Validates that walltime results exist and contain at least one benchmark.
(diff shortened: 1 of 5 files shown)


src/run/runner/wall_time/helpers.rs
```diff
@@ -41,9 +46,12 @@ pub fn validate_walltime_results(profile_folder: &Path) -> Result<()> {
...
-            )));
+            if !allow_empty {
+                bail!(add_empty_result_error_explanation(&format!(
+                    "No benchmarks found in walltime results file: {path:?}."
+                )));
+            }
+            debug!("No benchmarks found in {path:?} (allowed)");
         }
 
         found_valid_results = true;
```
adriencaccia: Since the code will bail if there is a parsing error, let's rename this variable to `found_results` so that it is not misleading

PR #164, opened 2025-11-24
feat: profile memory of command

not-matthias:
Changes in this PR: 
- Add MessagePack serialization for markers and heaptrack results
- Made the RunnerFifo shared across modes (walltime and memory use it at the moment)
- Added the heaptrack CLI + support to enable/disable it via IPC
- Added the memory executor (WIP)

TODO:
- [x] Create release CI for codspeed-heaptrack
- [x] Fix lints
- [x] Add upload implementation for memory
- [x] ~Parse the backend response~ skipped for now, we'll add this later
- [x] Ensure env forwarding works (cargo codspeed run)
- [x] Rename to memtrack

### DIFF
--- a/src/run/runner/tests.rs
+++ b/src/run/runner/tests.rs
@@ -130,23 +130,38 @@ async fn create_test_setup() -> (SystemInfo, RunData, TempDir) {
     (system_info, run_data, temp_dir)
 }
 
+// Uprobes set by memtrack, lead to crashes in valgrind because they work by setting breakpoints on the first
+// instruction. Valgrind doesn't rethrow those breakpoint exceptions, which makes the test crash.
+//
+// Therefore, we can only execute either valgrind or memtrack at any time, and not both at the same time.
+static BPF_INSTRUMENTATION_LOCK: OnceCell<Semaphore> = OnceCell::const_new();
+
+async fn acquire_bpf_instrumentation_lock() -> SemaphorePermit<'static> {
+    let semaphore = BPF_INSTRUMENTATION_LOCK
+        .get_or_init(|| async { Semaphore::new(1) })
+        .await;
+    semaphore.acquire().await.unwrap()
+}
+
 mod valgrind {
     use super::*;
 
-    async fn get_valgrind_executor() -> &'static ValgrindExecutor {
+    async fn get_valgrind_executor() -> (SemaphorePermit<'static>, &'static ValgrindExecutor) {
         static VALGRIND_EXECUTOR: OnceCell<ValgrindExecutor> = OnceCell::const_new();
 
-        VALGRIND_EXECUTOR
+        let executor = VALGRIND_EXECUTOR
             .get_or_init(|| async {
                 let executor = ValgrindExecutor;
                 let system_info = SystemInfo::new().unwrap();
                 executor.setup(&system_info, None).await.unwrap();
                 executor
             })
-            .await
+            .await;
+        let _lock = acquire_bpf_instrumentation_lock().await;
+
+        (_lock, executor)
     }
 
-    #[cfg(test)]
     fn valgrind_config(command: &str) -> Config {
         Config {
             mode: RunnerMode::Simulation,
@@ -159,7 +174,7 @@ mod valgrind {
     #[test_log::test(tokio::test)]
     async fn test_valgrind_executor(#[case] cmd: &str) {
         let (system_info, run_data, _temp_dir) = create_test_setup().await;
-        let executor = get_valgrind_executor().await;
+        let (_lock, executor) = get_valgrind_executor().await;
 
         let config = valgrind_config(cmd);
         executor
@@ -172,7 +187,7 @@ mod valgrind {
     #[test_log::test(tokio::test)]
     async fn test_valgrind_executor_with_env(#[case] env_case: (&str, &str)) {
         let (system_info, run_data, _temp_dir) = create_test_setup().await;
-        let executor = get_valgrind_executor().await;
+        let (_lock, executor) = get_valgrind_executor().await;
 
         let (env_var, env_value) = env_case;
         temp_env::async_with_vars(&[(env_var, Some(env_value))], async {
@@ -304,7 +319,11 @@ fi
 mod memory {
     use super::*;
 
-    async fn get_memory_executor() -> (SemaphorePermit<'static>, MemoryExecutor) {
+    async fn get_memory_executor() -> (
+        SemaphorePermit<'static>,
+        SemaphorePermit<'static>,
+        MemoryExecutor,
+    ) {
         static MEMORY_INIT: OnceCell<()> = OnceCell::const_new();
         static MEMORY_SEMAPHORE: OnceCell<Semaphore> = OnceCell::const_new();
 
@@ -321,10 +340,12 @@ mod memory {
             .await;
         let permit = semaphore.acquire().await.unwrap();
 
-        (permit, MemoryExecutor)
+        // Memory executor uses heaptrack which uses BPF-based instrumentation, which conflicts with valgrind.
+        let _lock = acquire_bpf_instrumentation_lock().await;
+
+        (permit, _lock, MemoryExecutor)
     }
 
-    #[cfg(test)]
     fn memory_config(command: &str) -> Config {
         Config {
             mode: RunnerMode::Memory,
@@ -337,7 +358,7 @@ mod memory {
     #[test_log::test(tokio::test)]
     async fn test_memory_executor(#[case] cmd: &str) {
         let (system_info, run_data, _temp_dir) = create_test_setup().await;
-        let (_permit, executor) = get_memory_executor().await;
+        let (_permit, _lock, executor) = get_memory_executor().await;
 
         let config = memory_config(cmd);
         executor
@@ -350,7 +371,7 @@ mod memory {
     #[test_log::test(tokio::test)]
     async fn test_memory_executor_with_env(#[case] env_case: (&str, &str)) {
         let (system_info, run_data, _temp_dir) = create_test_setup().await;
-        let (_permit, executor) = get_memory_executor().await;
+        let (_permit, _lock, executor) = get_memory_executor().await;
 
         let (env_var, env_value) = env_case;
         temp_env::async_with_vars(&[(env_var, Some(env_value))], async {


flake.nix
```diff
@@ -0,0 +1,51 @@
+{
```
GuillaumeLagrange: Let's not add flakes to our repos for now.
To share non trivial stuff between us, we can start small by having a dedicated codspeed-flakes private repo that we share, and use .envrc to do

`use flake ../path/to/flake/repo#repo-name-used-as-flake-name`

src/run/runner/shared/fifo.rs
```diff
@@ -0,0 +1,224 @@
...
+        &self.ctl_path
+    }
+
+    pub fn ack_path(&self) -> &Path {
+        &self.ack_path
+    }
+}
+
+// TODO: Find a better name for this (SharedBenchmarkData? BenchmarkData?)
+pub struct FifoData {
```
GuillaumeLagrange: `FifoBenchmarkData` is the best I have rn

src/run/runner/shared/fifo.rs
```diff
@@ -0,0 +1,224 @@
...
+                    self.send_cmd(FifoCommand::Ack).await?;
+                    continue;
+                }
+                FifoCommand::AddMarker { marker, .. } => {
+                    markers.push(*marker);
+                    self.send_cmd(FifoCommand::Ack).await?;
+                    continue;
+                }
+                FifoCommand::SetVersion(protocol_version) => {
+                    if *protocol_version < runner_shared::fifo::CURRENT_PROTOCOL_VERSION {
```
GuillaumeLagrange: I find matching against a.cmp(b) more elegant
```
use std::cmp::Ordering;

fn main() {
    match 2.cmp(&2) {
        Ordering::Less => println!("Less than 2."),
        Ordering::Greater => println!("Greater than 2."),
        Ordering::Equal => println!("Equal to 2."),
    }
}
```

crates/runner-shared/src/benchmark_results.rs
```diff
@@ -0,0 +1,100 @@
+use log::debug;
+pub use prost::Message;
+
+// Include the generated protobuf code
+include!(concat!(
```
GuillaumeLagrange: This is a bit awkward, couldnt we generate the rust file and use it as a normal rust file ?
And generation could be part of the crate's build.rs

crates/runner-shared/src/benchmark_results.rs
```diff
@@ -0,0 +1,100 @@
...
+    ) -> anyhow::Result<()> {
+        std::fs::create_dir_all(folder.as_ref())?;
+        std::fs::write(folder.as_ref().join(filename), self.encode_to_vec())?;
+
+        debug!("Saved {} result to {:?}", Self::name(), folder.as_ref());
+        Ok(())
+    }
+
+    fn save_to<P: AsRef<std::path::Path>>(&self, folder: P) -> anyhow::Result<()> {
+        self.save_file_to(folder, &format!("{}.binpb", Self::name()))
```
GuillaumeLagrange: Is binbp a standard of some sort? I couldnt find info about this :thinking:

src/run/mod.rs
```diff
@@ -148,6 +148,7 @@ pub enum RunnerMode {
     Instrumentation,
     Simulation,
     Walltime,
+    Memory,
```
GuillaumeLagrange, line 151: Since we're going to panic if we chose this, is there a way to hide this from the `--help` until we're ready ?

crates/runner-shared/Cargo.toml
```diff
@@ -14,3 +14,4 @@ prost = "0.14.1"
 
 [build-dependencies]
 prost-build = "0.14.1"
+protoc-bin-vendored = "3.2.0"
```
GuillaumeLagrange: what's the point of this ? Do we still need the libproto in the ci witht his ?

GuillaumeLagrange (changes requested):
CI does not build, and the CI change commit scares me a bit, going forward not gonna lie

Just so I know, have you had the protobuf decision challenged ?
Could be good to have a small design document just to make sure the whole team is aligned behind the choice.

crates/runner-shared/src/benchmark_results.rs
not-matthias: It's recommended by protbuf, but we can also use our own: https://protobuf.dev/programming-guides/techniques/#suffixes

crates/runner-shared/Cargo.toml
not-matthias: I added it to avoid having to install protoc in the monorepo PR, but I'll make sure to remove it from the CI here again as well!

crates/runner-shared/src/benchmark_results.rs
not-matthias: Which part do you find awkward? 

I previously tried to have a version that used the Rust structs, but it was much less readable imo. It would look something like this: 
```rust
#[derive(Clone, PartialEq, prost::Message)]
pub struct MetricList {
    #[prost(message, repeated, tag = "1")]
    pub metrics: Vec<Metric>,
}

#[derive(Clone, PartialEq, prost::Message)]
pub struct Metric {
    #[prost(oneof = "metric::Kind", tags = "10")]
    pub kind: Option<metric::Kind>,
}

pub mod metric {
    #[derive(Clone, PartialEq, prost::Oneof)]
    pub enum Kind {
        #[prost(message, tag = "10")]
        Heaptrack(super::HeaptrackMetric),
        // Add other metrics here
    }
}

#[derive(Clone, PartialEq, prost::Message)]
pub struct HeaptrackMetric {
    #[prost(message, repeated, tag = "1")]
    pub events: Vec<HeaptrackEvent>,
}

#[derive(Clone, PartialEq, prost::Message)]
pub struct HeaptrackEvent {
    #[prost(uint32, tag = "1")]
    pub pid: u32,
    #[prost(uint32, tag = "2")]
    pub tid: u32,
    #[prost(uint64, tag = "3")]
    pub timestamp: u64,
    #[prost(enumeration = "heaptrack_event::HeaptrackEventType", tag = "4")]
    pub r#type: i32,
    #[prost(uint64, tag = "5")]
    pub addr: u64,
    #[prost(uint64, optional, tag = "6")]
    pub size: Option<u64>, // Not used for FREE
}
```

src/run/runner/shared/fifo.rs
not-matthias: Ohhh, that does indeed look much better. Good catch!

src/run/mod.rs
not-matthias: I removed the panic upon upload. I think it's fine if we keep it as-is, since you can't enable it yet in the action.

crates/runner-shared/src/benchmark_results.rs
```diff
@@ -0,0 +1,83 @@
...
+    Munmap { size: u64 },
+    Brk { size: u64 },
+}
+
+pub trait BenchmarkResultExt
+where
+    Self: Sized + Serialize,
+{
+    fn name() -> &'static str {
+        // WARN: This doesn't support generic types
```
GuillaumeLagrange: This belong in the docstring, then each struct will be free to implement an override for this if they need to handle generic types

crates/runner-shared/src/benchmark_results.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    fn save_to<P: AsRef<std::path::Path>>(&self, folder: P) -> anyhow::Result<()> {
+        self.save_file_to(folder, &format!("{}.msgpack", Self::name()))
+    }
+
+    fn save_with_pid_to<P: AsRef<std::path::Path>>(
+        &self,
+        folder: P,
+        pid: u32,
```
GuillaumeLagrange: use `pid_t`

crates/runner-shared/src/benchmark_results.rs
```diff
@@ -0,0 +1,83 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct HeaptrackResult {
+    pub events: Vec<HeaptrackEvent>,
+}
+impl BenchmarkResultExt for HeaptrackResult {}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct HeaptrackEvent {
+    pub pid: u32,
```
GuillaumeLagrange: use `pid_t`

GuillaumeLagrange: I think it's finally time to export a `pid_t` from runner-shared and use it from here

crates/runner-shared/src/benchmark_results.rs
```diff
@@ -0,0 +1,83 @@
+use crate::fifo::MarkerType;
+use log::debug;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MarkerResult {
```
GuillaumeLagrange: I get the initial idea, but `MarkerResult` does feel a bit weird, especially since this struct has both the markers and the uri by timestamp.

How about `ExecutionTimestamps` or something like this ?

crates/runner-shared/src/benchmark_results.rs
```diff
@@ -0,0 +1,83 @@
...
+    Free,
+    Realloc { size: u64 },
+    Calloc { size: u64 },
+    AlignedAlloc { size: u64 },
+    Mmap { size: u64 },
+    Munmap { size: u64 },
+    Brk { size: u64 },
+}
+
+pub trait BenchmarkResultExt
```
GuillaumeLagrange: I think what we are defining here is "A benchmark execution output that we persist to file"

I like `Artifact` to convey the idea of persisting this to file rather than `Result`, WDYT ?
And I think `Execution` rather than `Benchmark` is more neutral, and technically this is the output of executor.

Since we are persisting these let's try and nail the naming before releasing

crates/runner-shared/src/fifo.rs
```diff
@@ -37,6 +37,7 @@ pub enum Command {
     Ack,
     #[deprecated(note = "Use `GetIntegrationMode` instead")]
     PingPerf,
+    #[deprecated(note = "Not needed anymore, use `integration` from WalltimeResults instead")]
```
GuillaumeLagrange: If we want to deprecate this, we also need to rename `WalltimeResults` to something more generic like `ExecutionResults` or `BenchmarkResults`, because it's no longer linked to the walltime executor directly

Cargo.toml
```diff
@@ -76,7 +76,7 @@ rstest_reuse = "0.7.0"
 shell-quote = "0.7.2"
 
 [workspace]
-members = ["crates/runner-shared"]
+members = ["crates/*"]
```
GuillaumeLagrange: personal preference, but I prefer explicit crate declaration rather than glob, pains the picture of the repo better when landing in a repo and looking into the Cargo.toml

crates/heaptrack/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+[lib]
+name = "heaptrack"
+path = "src/lib.rs"
+
+[[bin]]
+name = "codspeed-heaptrack"
+path = "src/main.rs"
+
+[dependencies]
+anyhow = "1.0"
```
GuillaumeLagrange: a lot of these should be workspace dependencies unless we have a good reason for version mismatch

crates/heaptrack/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+serde = { version = "1.0.228", features = ["derive"] }
+runner-shared = { path = "../runner-shared" }
+ipc-channel = "0.18"
+static_assertions = "1.1"
+itertools = "0.14.0"
+
+[build-dependencies]
+libbpf-cargo = "0.25.0"
+bindgen = "0.71"
+vmlinux = { git = "https://github.com/libbpf/vmlinux.h.git" }
```
GuillaumeLagrange: This seem quite unstable according to their readme, let's at least pin the rev

crates/memtrack/build.rs
```diff
@@ -0,0 +1,32 @@
...
+        .source("src/bpf/heaptrack.bpf.c")
+        .clang_args([
+            "-I",
+            &vmlinux::include_path_root().join(arch).to_string_lossy(),
+        ])
+        .build_and_generate(&heaptrack_out)
+        .unwrap();
+
+    // Generate bindings for event.h
+    let bindings = bindgen::Builder::default()
```
GuillaumeLagrange, line 24: We moved away from bindgen for codspeed's core library, what makes us confident that using bindgen here will not be a repeat of this ?

crates/heaptrack/src/ebpf/memtrack.rs
```diff
@@ -0,0 +1,274 @@
...
+            .context(format!(
+                "Failed to attach realloc uretprobe in {}",
+                libc_path.display()
+            ))?;
+        self.probes.push(link);
+
+        Ok(())
+    }
+
+    pub fn attach_aligned_alloc(&mut self, libc_path: &Path) -> Result<()> {
```
GuillaumeLagrange: overall there is a lot of repetition in this file. Can't we extract the key parts for each function we want to attach and factorize the common code ?

crates/memtrack/build.rs
```diff
@@ -0,0 +1,32 @@
...
+    let arch = env::var("CARGO_CFG_TARGET_ARCH")
+        .expect("CARGO_CFG_TARGET_ARCH must be set in build script");
+    let out_dir =
+        PathBuf::from(env::var_os("OUT_DIR").expect("OUT_DIR must be set in build script"));
+    let heaptrack_out = out_dir.join("heaptrack.skel.rs");
+    SkeletonBuilder::new()
+        .source("src/bpf/heaptrack.bpf.c")
+        .clang_args([
+            "-I",
+            &vmlinux::include_path_root().join(arch).to_string_lossy(),
```
GuillaumeLagrange, line 18: If I understand correctly, this builds using bpf headers for the current kernel. Wont that be a huge issue going forward that basically make shipping pre-built heaptrack is going to be impossible ?

crates/heaptrack/src/ebpf/c/heaptrack.bpf.c
```diff
@@ -0,0 +1,491 @@
...
+/* Map to store calloc parameters between entry and return */
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 10000);
+    __type(key, __u64);   /* tid */
+    __type(value, __u64); /* nmemb * size (total size) */
+} calloc_size SEC(".maps");
+
+SEC("uprobe")
+int uprobe_calloc(struct pt_regs *ctx) {
```
GuillaumeLagrange: same for this file, if I understand correctly it's not generated, we could make this much less repetitive by using proc macros. WDYT ?

This would also ease integration with custom allocator in the future as we could expose the proc macros and clear and documented arguments to each, and the burden of itnegration would be quite small IMO

crates/heaptrack/test.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+# Build tests and capture the integration test binary path from cargo output
+output=$(cargo test --release --no-run 2>&1)
```
GuillaumeLagrange: is `jq` available in CI ? Case we could use `--message-format=json | jq` to have something a bit more robust

crates/heaptrack/test.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+# Build tests and capture the integration test binary path from cargo output
```
GuillaumeLagrange: There needs to be an explanation as to why we have a dedicated test shell script rather than using the usual cargo test

crates/memtrack/src/lib.rs
```diff
@@ -1,14 +1,24 @@
+#[cfg(feature = "ebpf")]
```
GuillaumeLagrange, line 1: would be easier to have a root ebpf module that we re-export rather than specifying the bpf feature on all modules

crates/memtrack/build.rs
not-matthias: We control the build process and then distribute the binaries, whereas in the core library we distributed the crate/source code which lead to issues when building on all kinds of architectures.

crates/heaptrack/test.sh
not-matthias: Turns out we can can run the tests with `sudo -E` which forwards all the environment variables (see: https://github.com/rust-lang/cargo/issues/5999).

crates/heaptrack/src/ebpf/c/heaptrack.bpf.c
not-matthias: Did a big refactor. We now have helper functions and macros (`UPROBE_WITH_ARGS`) which make defining new uprobes much easier. Lmk if this aligns with you had in mind

crates/memtrack/build.rs
not-matthias: There's CO-RE (compile once, run everywhere) which is also supported by libbpf so this shouldn't be an issue: https://github.com/libbpf/libbpf?tab=readme-ov-file#bpf-co-re-compile-once--run-everywhere

But I'll look into it to ensure we actually build the binary with this.

crates/runner-shared/src/benchmark_results.rs
not-matthias: I'm still using libc's type for now

crates/memtrack/build.rs
not-matthias: Looks good, as long as the kernel is built with BTF support (Ubuntu 20+) then we're fine. Since the tests are already running on CI, this shouldn't be an issue.

src/run/runner/shared/fifo.rs
```diff
@@ -0,0 +1,224 @@
...
+        let mut len_buffer = [0u8; 4];
+        self.ctl_fifo.read_exact(&mut len_buffer).await?;
+        let message_len = u32::from_le_bytes(len_buffer) as usize;
+
+        let mut buffer = vec![0u8; message_len];
+        loop {
+            if self.ctl_fifo.read_exact(&mut buffer).await.is_ok() {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
not-matthias: we'll already catch this waaaay ahead of time in the CI since it's caused by the integration.

src/run/runner/shared/fifo.rs
```diff
@@ -0,0 +1,226 @@
...
+    // Create the FIFO with RWX permissions for the owner
+    nix::unistd::mkfifo(path.as_ref(), nix::sys::stat::Mode::S_IRWXU)?;
+
+    Ok(())
+}
+
+pub struct GenericFifo {
+    ctl_path: PathBuf,
+    ack_path: PathBuf,
+    sender: TokioPipeSender,
```
GuillaumeLagrange: for clarity, we could name these fields `ctl_sender` and `ack_receiver`, even if these are implementation details internal to this module, would make things a tiny bit clearer

src/run/runner/wall_time/perf/mod.rs
```diff
@@ -497,11 +435,19 @@ impl BenchmarkData {
...
+            markers: self.fifo_data.markers.clone(),
             debug_info_by_pid,
         };
         metadata.save_to(&path).unwrap();
 
         Ok(())
     }
 }
+
+impl Deref for BenchmarkData {
```
GuillaumeLagrange: For the foreseeable future, `BenchmarkData` will hold data other than the `FifoData`, I think it's better to stick to this kind of `Deref` implem when the struct is just a wrapper around a single field, and otherwise be more explicit.

If it's an easy way to implement encapsulation by keeping the field private, let's just add a `fifo_data` method

crates/runner-shared/src/artifacts.rs
```diff
@@ -0,0 +1,84 @@
...
+impl ArtifactExt for ExecutionTimestamps {}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MemtrackArtifact {
+    pub events: Vec<MemtrackEvent>,
+}
+impl ArtifactExt for MemtrackArtifact {}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct MemtrackEvent {
```
GuillaumeLagrange: I think we could benefit from one module per instrument here.

In the future, we'll be able to toggle some of the events through features to make sure each instrument only has access to its events/artifact structures in runner_shared

src/run/runner/shared/fifo.rs
```diff
@@ -129,7 +127,7 @@ impl RunnerFifo {
         &mut self,
         mut health_check: impl AsyncFnMut() -> anyhow::Result<bool>,
         mut handle_cmd: impl AsyncFnMut(&FifoCommand) -> anyhow::Result<FifoCommand>,
-    ) -> anyhow::Result<FifoBenchmarkData> {
+    ) -> anyhow::Result<(MarkerResult, FifoBenchmarkData)> {
```
GuillaumeLagrange: I think some of the naming was not properly fixed up here, FYI.

Don't necessarily bother redoing everything tho

src/run/runner/shared/fifo.rs
```diff
@@ -127,7 +127,7 @@ impl RunnerFifo {
         &mut self,
         mut health_check: impl AsyncFnMut() -> anyhow::Result<bool>,
         mut handle_cmd: impl AsyncFnMut(&FifoCommand) -> anyhow::Result<FifoCommand>,
-    ) -> anyhow::Result<(MarkerResult, FifoBenchmarkData)> {
+    ) -> anyhow::Result<(ExecutionTimestamps, FifoBenchmarkData)> {
```
GuillaumeLagrange, line 130: here they are!

src/run/runner/tests.rs
```diff
@@ -346,21 +367,21 @@ mod memory {
...
-            let cmd = env_var_validation_script(env_var, env_value);
-            let config = memory_config(&cmd);
-            executor
-                .run(&config, &system_info, &run_data, &None)
-                .await
-                .unwrap();
-        })
-        .await;
-    }
+    // #[apply(env_test_cases)]
```
GuillaumeLagrange: Is this commented on purpose ?

not-matthias: ah, no that shouldn't be commented out. only needed it for debugging the failing tests. i'll fix it 👍

src/run/runner/shared/fifo.rs
not-matthias: Changed this to `ctl_sender` and `ack_reader` (since the structs are also using sender/reader).

PR #165, opened 2025-11-25
Feat: add the exec subcommand to run an arbitrary rust binary

GuillaumeLagrange:
- Depends on the monorepo pr for the new endpoint
- Refactor to move out of the `run` module the whole executor logic
- Create a new `ExecutionContext` that holds all the data required by the executor and other parts to work
- Factorize the logic between `exec` and `run` while keeping some stuff modular
- Parse the perf map during wt teardown if not outputing pipedata
   - This is temporary, while the upstream linux-perf-data cannot parse the `pipedata` format, ideally we remove the `/proc/<pid>/maps` parsing logic altogether

Some of the logic would still need to be refactored to be perfectly clean, but the PR is starting to be quite huge already

CI fails because of unrelated version mismatch within the repo (the plan job), should not affect review

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -691,7 +691,7 @@ dependencies = [
 
 [[package]]
 name = "exec-harness"
-version = "4.4.1"
+version = "4.4.2-beta.1"
 dependencies = [
  "anyhow",
  "clap",
@@ -1572,7 +1572,7 @@ dependencies = [
 
 [[package]]
 name = "memtrack"
-version = "4.4.1"
+version = "4.4.2-beta.1"
 dependencies = [
  "anyhow",
  "bindgen",
--- a/crates/exec-harness/Cargo.toml
+++ b/crates/exec-harness/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "exec-harness"
-version = "4.4.1"
+version = "4.4.2-beta.1"
 edition = "2024"
 
 [dependencies]
--- a/crates/memtrack/Cargo.toml
+++ b/crates/memtrack/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "memtrack"
-version = "4.4.1"
+version = "4.4.2-beta.1"
 edition = "2024"
 repository = "https://github.com/CodSpeedHQ/runner"
 publish = false
@@ -16,11 +16,7 @@ required-features = ["ebpf"]
 
 [features]
 default = ["ebpf"]
-ebpf = [
-    "dep:libbpf-rs",
-    "dep:libbpf-cargo",
-    "dep:vmlinux",
-]
+ebpf = ["dep:libbpf-rs", "dep:libbpf-cargo", "dep:vmlinux"]
 
 [dependencies]
 anyhow = { workspace = true }


crates/exec-harness/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+#[derive(Parser, Debug)]
+#[command(name = "exec-harness")]
+#[command(about = "CodSpeed exec harness - wraps commands with performance instrumentation")]
+struct Args {
+    /// Optional benchmark name (defaults to command filename)
+    #[arg(long)]
+    name: Option<String>,
+
+    /// The command and arguments to execute
+    command: Vec<String>,
```
not-matthias, line 22: TODO: Does this work when passing a string like this? `exec-harness -- ls -alh` -> do we then have a vec of `vec!["ls", "-alh"]`?

not-matthias: If it's just a single string, we also wouldn't need the is_empty check in main

crates/exec-harness/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // Derive benchmark name from command if not provided
+    let bench_name = args.name.unwrap_or_else(|| {
+        // Extract filename from command path
+        let cmd = &args.command[0];
+        std::path::Path::new(cmd)
+            .file_name()
+            .and_then(|n| n.to_str())
+            .map(|s| s.to_string())
```
not-matthias: This could just be `to_string_lossy`

crates/exec-harness/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+fn main() {
+    let args = Args::parse();
+
+    if args.command.is_empty() {
+        eprintln!("Error: No command provided");
+        process::exit(1);
+    }
+
+    // Derive benchmark name from command if not provided
+    let bench_name = args.name.unwrap_or_else(|| {
```
not-matthias, line 33: Can be simplified to `.or_else(...).unwrap_or("exec-harness".into())` to reduce nesting

crates/exec-harness/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+        .unwrap();
+
+    const NUM_ITERATIONS: usize = 1;
+    let mut times_per_round_ns = Vec::with_capacity(NUM_ITERATIONS);
+
+    hooks.start_benchmark().unwrap();
+    for _ in 0..NUM_ITERATIONS {
+        // Spawn the command
+        let mut child = match process::Command::new(&args.command[0])
+            .args(&args.command[1..])
```
not-matthias: Does this work with all kinds of bash expansions? e.g. what if i want to run `rg src/*.cpp`, will that also work?

We had this issue before where `*` wasn't expanded properly, so maybe it makes sense to only accept a `String` cmd which is then executed using `bash -c {cmd}`

crates/exec-harness/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// The command and arguments to execute
+    command: Vec<String>,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    if args.command.is_empty() {
+        eprintln!("Error: No command provided");
+        process::exit(1);
```
not-matthias: Let's use anyhow in the exec-harness, then we can replace this with
```
bail!("Error: No command provided");
```

This will also improve and simplify the error handling code down below quite a lot: 
```rust
 let status = match child.wait() {
            Ok(status) => status,
            Err(e) => {
                eprintln!("Failed to wait for command: {e}");
                process::exit(1);
            }
        };
```
becomes
```rust
 let status = child.wait().context("Failed to wait for command")?;
```

crates/exec-harness/src/walltime.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+impl WalltimeResults {
+    pub fn from_benchmarks(benchmarks: Vec<WalltimeBenchmark>) -> Result<Self> {
+        Ok(WalltimeResults {
+            instrument: Instrument {
+                type_: "walltime".to_string(),
+            },
+            creator: Creator {
+                name: "codspeed-rust".to_string(),
```
not-matthias, line 36: please also add a todo here to not impersonate codspeed-rust, so that we don't forget

src/executor/wall_time/perf/perf_map.rs
```diff
@@ -79,7 +79,30 @@ impl ModuleSymbols {
...
+                    const PAGE_SIZE: u64 = 4096;
+                    let addr = symbols[i].addr;
+                    let end_addr = addr.next_multiple_of(PAGE_SIZE) + PAGE_SIZE;
+                    symbols[i].size = end_addr.saturating_sub(addr);
+                }
+            }
+        }
+
+        // Filter out any symbols that still have zero size
+        symbols.retain(|symbol| symbol.size > 0);
```
not-matthias, line 104: Remove this, since we modify the size in the loop above.

src/exec/mod.rs
```diff
@@ -23,51 +25,40 @@ pub async fn run(
...
-        perf_unwinding_mode: args.shared.perf_run_args.perf_unwinding_mode,
-        profile_folder: args.shared.profile_folder,
-        skip_upload: args.shared.skip_upload,
-        skip_run: args.shared.skip_run,
-        skip_setup: args.shared.skip_setup,
-        allow_empty: args.shared.allow_empty,
-    };
+        if execution_context.is_local() {
+            poll_results::poll_results(api_client, upload_result.run_id).await?;
+            end_group!();
```
not-matthias: we have a duplicate `end_group` here I think

src/exec/poll_results.rs
```diff
@@ -0,0 +1,89 @@
...
+    let table_rows: Vec<BenchmarkRow> = results
+        .iter()
+        .map(|result| BenchmarkRow {
+            name: result.benchmark.name.clone(),
+            time: helpers::format_duration(result.time, Some(2)),
+        })
+        .collect();
+
+    Table::new(&table_rows).with(Style::modern()).to_string()
+}
```
not-matthias: Can we share this logic with the other `poll_results` implementation?

not-matthias: same for the constants above

not-matthias: or maybe use a single implementation for both (e.g. based on the executor), since the only thing that differs is the request and the response struct type. 

I'll have to also implement something like this for the memory executor, so we could abstract it.

src/executor/mod.rs
```diff
@@ -130,48 +122,22 @@ pub async fn execute_benchmarks(
...
-                api_client,
-                &execution_context.provider,
-                upload_result.run_id,
-                output_json,
-            )
-            .await?;
-            end_group!();
-        }
-    }
-
```
not-matthias: is there any reason you removed this from here? We will almost always have the same logic for uploading/skipping the upload, so we could pass the executor type around.

src/run/mod.rs
```diff
@@ -184,16 +184,50 @@ pub async fn run(
...
+            poll_results::poll_results(
+                api_client,
+                &execution_context.provider,
+                upload_result.run_id,
+                output_json,
+            )
+            .await?;
+            end_group!();
+        }
+    }
```
not-matthias: as mentioned above, we now have duplicate code here and in the exec/mod.rs, which we could try to avoid. If there's no other way, we can keep it like this but let's make a conscious decision. 

We already have the shared `execute_benchmarks` function, so it would be another abstraction above that.

src/executor/wall_time/executor.rs
```diff
@@ -94,9 +94,9 @@ pub struct WallTimeExecutor {
 }
 
 impl WallTimeExecutor {
-    pub fn new() -> Self {
+    pub fn new(output_pipedata: bool) -> Self {
```
not-matthias: Can we keep this as `new() -> Self` and add a new function `new_with_option(bool) -> Self` (feel free to use a better name) that allows specifying the output pipedata argument. 

The main benefits are that we don't have to add the default value `true` to the tests and many other places. This would make removing this flag much easier later on, since we only have to change it in the one place that actually needs it.

src/executor/wall_time/perf/mod.rs
```diff
@@ -317,13 +270,15 @@ impl PerfRunner {
 
             match cmd {
                 FifoCommand::CurrentBenchmark { pid, uri } => {
-                    bench_order_by_timestamp.push((current_time(), uri));
+                    bench_order_by_timestamp.push((current_time(), uri.clone()));
```
not-matthias: We don't need the clone here

src/executor/wall_time/perf/parse_perf_file.rs
```diff
@@ -0,0 +1,122 @@
...
+            );
+            continue;
+        }
+
+        if record_path_string.starts_with("[") && record_path_string.ends_with("]") {
+            // Skip special mappings
+            debug!(
+                "Skipping special mapping: {} - {:x}-{:x}",
+                record_path_string, record.address, end_addr
+            );
```
not-matthias: Have you checked how many of those special mappings we have? Maybe it's better to make this also `trace!` to not spam the logs.

src/executor/wall_time/perf/parse_perf_file.rs
```diff
@@ -0,0 +1,122 @@
...
+    let mut symbols_by_pid = HashMap::<pid_t, ProcessSymbols>::new();
+    let mut unwind_data_by_pid = HashMap::<pid_t, Vec<UnwindData>>::new();
+
+    // TODO: Handle pipedata here once linux_perf_data supports it upstream
+    let PerfFileReader {
+        mut perf_file,
+        mut record_iter,
+    } = PerfFileReader::parse_file(reader)?;
+
+    const PROT_EXEC: u32 = 0x4;
```
not-matthias: Let's use `libc::PROT_EXEC` here since we already have that as a dep

src/executor/wall_time/perf/parse_perf_file.rs
GuillaumeLagrange: It's usually 4 our 5 per pid approximately, but yeah I'll just move this to trace

crates/exec-harness/src/main.rs
GuillaumeLagrange: bash extensions are out of scope for now, so are piping etc

The very first version only supports basic rust executables, and their cli, which should work with the current implem

GuillaumeLagrange: Although I do agree this will be reworked soon

crates/exec-harness/src/main.rs
GuillaumeLagrange: Right now it only works with exec-harness -- ls -- -alh because otherwise clap messes with argument parsing.

This is a bit out of scope of the first version, although I completely agree it should be handled

src/exec/poll_results.rs
GuillaumeLagrange: SOrry for this, actually this is done in a later commit

src/executor/wall_time/perf/perf_map.rs
GuillaumeLagrange: Actually even with this change, there are still a lot of 0 sized symbols here. Mostly because sometimes multiple symbols are at the same address.
I figure it's safer to filter out stuff that has not be caught by the above logic in the first place here, LMK if that's not ok with you

src/run/mod.rs
GuillaumeLagrange: changed this to take a closure for polling as an argument to `execute_benchmark` adn moving back the logic to the shared `execute_benchmarks` function

src/executor/wall_time/executor.rs
GuillaumeLagrange: I've just kept it for the PerfRunner inside since it's really local to the walltime executor

crates/exec-harness/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// The command and arguments to execute
+    command: Vec<String>,
+}
+
+fn main() -> Result<()> {
+    let args = Args::parse();
+
+    if args.command.is_empty() {
+        eprintln!("Error: No command provided");
+        process::exit(1);
```
not-matthias: I think you missed this one, it could also be rewritten with `bail!` (unless there's a downside of doing so that I'm missing 🤔 )

PR #169, opened 2025-12-11
fix: forward entire system env to sudo

not-matthias:


src/executor/helpers/env.rs
```diff
@@ -15,7 +15,7 @@ pub fn get_base_injected_env(
         #[allow(deprecated)]
         RunnerMode::Instrumentation | RunnerMode::Simulation => "instrumentation",
         RunnerMode::Walltime => "walltime",
-        RunnerMode::Memory => "memory",
+        RunnerMode::Memory => "analysis",
```
GuillaumeLagrange: I'm a bit confused here, if we have an "analysis" mode here, why don't we push it for Simulation and Instrumentation as well ? Is it something that we do not do yet because we want to maintain some form of backwards compatibility ?

Overall, it's still a bit unclear what the `runner mode` and `integration mode` are, I believed runner mode would stay the same and we would decouple this from integration mode. To me, `analysis` was an integration mode.

src/executor/helpers/run_with_env.rs
```diff
@@ -0,0 +1,65 @@
...
+        "source {} && {}",
+        env_file.path().display(),
+        original_command
+    );
+    cmd_builder.wrap("bash", ["-c", &bash_command]);
+
+    Ok((cmd_builder, env_file))
+}
+
+pub fn create_env_file(extra_env: &HashMap<&'static str, String>) -> Result<NamedTempFile> {
```
GuillaumeLagrange: does not need to be pub

GuillaumeLagrange (approved):
olgtm

GuillaumeLagrange (changes requested):
sorry did not mean to approve in the previous review, let me know if you need help rebasing i've merged #165

src/executor/helpers/env.rs
not-matthias: > I'm a bit confused here, if we have an "analysis" mode here, why don't we push it for Simulation and Instrumentation as well ? Is it something that we do not do yet because we want to maintain some form of backwards compatibility ?

`CODSPEED_RUNNER_MODE` is used by cargo-codspeed: https://github.com/CodSpeedHQ/codspeed-rust/blob/9181bce1e3468a7387175b0f2b8e92f432158def/crates/cargo-codspeed/src/app.rs#L15-L18

If we would pass `analysis` for simulation/instrumentation then we would have to ensure that the `cargo-codspeed` version supports it. 


> Overall, it's still a bit unclear what the `runner mode` and `integration mode` are, I believed runner mode would stay the same and we would decouple this from integration mode. To me, `analysis` was an integration mode.

Yes, I agree. Unfortunately we can't really rename the env variable for backwards compatibility reasons

not-matthias: As discussed together, pass `memory` for now and update the logic in cargo-codspeed.

PR #171, opened 2025-12-15
Manage installation and release of the exec harness and codspeed-memtrack

GuillaumeLagrange:


src/binary_installer/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+        .output()
+        .context("Failed to execute installer command")?;
+
+    if !output.status.success() {
+        bail!(
+            "Failed to install {binary_name} version {version}. Installer exited with output: {output:?}",
+        );
+    }
+
+    info!("Successfully installed {binary_name} version {version}");
```
adriencaccia, line 64: Let's check that the command is correctly installed and available is in the path here after installing it

PR #177, opened 2025-12-19
ci: do not trigger action workflow if we are not releasing the runner

GuillaumeLagrange:


art049:
Is this still needed with #180 ? @GuillaumeLagrange

PR #180, opened 2025-12-22
feat: update release flow to make sure only runner releases are marked as latest

GuillaumeLagrange:
I did not want to over-engineer something for the tag, so I just made it take the latest tag in `scripts/release.sh`, which gives the limitation of only working with one crate per script invocation.

Tested here
https://github.com/CodSpeedHQ/runner/actions/runs/20457122332
which created fake release here
https://github.com/CodSpeedHQ/runner/releases/tag/memtrack-v1.0.0-alpha.13
image
(Screenshot since the release will most likely be removed after merge to clean things up)

### DIFF
--- a/scripts/cleanup_alpha.sh
+++ b/scripts/cleanup_alpha.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+set -e
+
+echo "=== Alpha Release Cleanup Script ==="
+echo ""
+
+# Fetch all alpha releases
+echo "Fetching alpha releases..."
+RELEASES=$(gh release list --limit 1000 | grep "alpha" | awk -F'\t' '{print $3}' || true)
+
+# Fetch all alpha tags
+echo "Fetching alpha tags..."
+git fetch --tags 2>/dev/null || true
+TAGS=$(git tag -l "*alpha*" || true)
+
+# Combine and get unique names
+ALL_ITEMS=$(echo -e "$RELEASES\n$TAGS" | sort -u | grep -v '^$' || true)
+
+echo ""
+echo "========================================="
+echo "FOUND ITEMS:"
+echo "========================================="
+echo ""
+
+if [ -z "$ALL_ITEMS" ]; then
+    echo "No alpha releases or tags found."
+    exit 0
+fi
+
+# Display each item with its status
+for ITEM in $ALL_ITEMS; do
+    HAS_RELEASE=""
+    HAS_TAG=""
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -16,7 +16,7 @@ This repository is a Cargo workspace containing multiple crates. The release pro
 For any crate other than the main runner:
 
 ```bash
-cargo release -p <PACKAGE_NAME> --execute <VERSION_BUMP>
```
art049, line 19: We should avoid as much as possible having an extra script. Cargo release should do already everything we need. then the dist ci workflow should handle releases creation/publish

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - uses: actions/checkout@v4
+
+      - name: Create draft release
+        run: |
+          RELEASE_TAG=$(echo ${PLAN} | jq -r '.announcement_tag')
+
+          echo "Creating draft release ${RELEASE_TAG}"
+          gh release create "$RELEASE_TAG" --draft --notes "" --title "Release $RELEASE_TAG"
+          echo "Release created successfully"
```
art049, line 39: Can't we create the release with the proper details upfront?

.github/workflows/post-announce.yml
```diff
@@ -0,0 +1,60 @@
...
+          echo "new_version=${NEW_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "release_tag=${RELEASE_TAG}" >> "$GITHUB_OUTPUT"
+
+          if [ "${IS_PRE_RELEASE}" == "true" ]; then
+            echo "runner_released=false" >> "$GITHUB_OUTPUT"
+            echo "Pre-release detected, will not mark as latest or bump action"
+          elif [ -z "${NEW_VERSION}" ] || [ "${NEW_VERSION}" == "null" ]; then
+            echo "runner_released=false" >> "$GITHUB_OUTPUT"
+            echo "No codspeed-runner release found in plan"
+          else
```
art049, line 38: I think in this case, we still need to publish the release but without putting it as latest. WDYT?

GuillaumeLagrange: the publishing part is already done by dist here I'll update the wording to make it clearer

 https://github.com/CodSpeedHQ/runner/blob/cod-1894-ensure-only-runner-releases-can-be-marked-as-latest-release/.github/workflows/release.yml

PR #185, opened 2026-01-05
Add walltime benchmarking algorithm to exec-harness

GuillaumeLagrange:


crates/exec-harness/src/walltime/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+    bench_uri: String,
+    command: Vec<String>,
+    config: &ExecutionOptions,
+) -> Result<Vec<u128>> {
+    let warmup_time_ns = config.warmup_time_ns;
+    let hooks = InstrumentHooks::instance();
+
+    let iterations_to_perform = if warmup_time_ns > 0 {
+        let warmup_start_ts_ns = InstrumentHooks::current_timestamp();
+        let warmup_end_ts = warmup_start_ts_ns + warmup_time_ns;
```
not-matthias: I was a bit confused why we already compute the end timestamp here, before even executing the loop. Since it's only used in the loop header, we could write `while InstrumentHooks::current_timestamp() < warmup_start_ts_ns + warmup_time_ns {`

crates/exec-harness/src/walltime/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+        let bench_iteration_start_ts_ns = InstrumentHooks::current_timestamp();
+        let status = child
+            .wait()
+            .context("Failed to wait for command to finish")?;
+
+        let bench_iteration_end_ts_ns = InstrumentHooks::current_timestamp();
+        hooks.add_benchmark_timestamps(bench_iteration_start_ts_ns, bench_iteration_end_ts_ns);
+
+        if !status.success() {
+            bail!("Command exited with non-zero status: {status}");
```
not-matthias: This could be shared in a closure to reduce code duplication

crates/exec-harness/src/walltime/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+        hooks.start_benchmark().unwrap();
+        while InstrumentHooks::current_timestamp() < warmup_end_ts {
+            let warmup_iteration_start_ts_ns = InstrumentHooks::current_timestamp();
+            let status = Command::new(&command[0])
+                .args(&command[1..])
+                .status()
+                .context("Failed to execute command")?;
+            let warmup_iteration_end_ts_ns = InstrumentHooks::current_timestamp();
+            hooks
+                .add_benchmark_timestamps(warmup_iteration_start_ts_ns, warmup_iteration_end_ts_ns);
```
not-matthias: We're including the warmup timestamps in the actual benchmark run.

We currently need to ensure that the timestamps actually only track the benchmark execution, since we use it to calculate the approximate time per frame in the flamegraph. Maybe instead save the time, then submit it in the check below that exits after 1 warmup iteration.

crates/exec-harness/tests/integration_test.rs
```diff
@@ -0,0 +1,326 @@
...
+        &exec_opts,
+    )?;
+
+    // Should have run at least 1 time, but not an excessive amount
+    assert!(!times.is_empty(), "Expected at least 1 iteration");
+    assert!(
+        times.len() < 1000,
+        "Expected fewer than 1000 iterations due to max_time constraint, got {}",
+        times.len()
+    );
```
not-matthias, line 93: There's a good chance that this will fail on faster machines since `echo "test"` might run faster. Instead, you could run `sleep 0.01` which will be deterministic across machines.

crates/exec-harness/src/walltime/mod.rs
not-matthias: You could always append both start and stop timestamps to `times_per_round_ns`, then when the warmup is done you can easily discard the values with `.clear()` ([that's how Go does it](https://github.com/golang/go/blob/9b2e3b9a02bd8872bdbf5c6086674fa6b4bc8ef9/src/testing/benchmark.go#L166-L183))

crates/exec-harness/src/walltime/mod.rs
GuillaumeLagrange: actually it's even more scuffed than that fo rnow since the profiling of only one iteration will be shown.

For now, I'll leave it like that, and we'll revisit when we update the profile to manage all the pids

crates/runner-shared/src/walltime_results.rs
```diff
@@ -44,16 +44,123 @@ pub struct WalltimeBenchmark {
...
+            stats: BenchmarkStats {
+                min_ns,
+                max_ns,
+                mean_ns,
+                stdev_ns,
+                q1_ns,
+                median_ns,
+                q3_ns,
+                rounds,
+                total_time,
```
adriencaccia: As in all the existing implementations, this should be in seconds, not in ns. Let's add a `///` doc on the struct field definition to explain that.

crates/runner-shared/src/walltime_results.rs
```diff
@@ -63,3 +170,47 @@ pub struct WalltimeResults {
...
+
+impl WalltimeResults {
+    pub fn from_benchmarks(benchmarks: Vec<WalltimeBenchmark>) -> anyhow::Result<Self> {
+        Ok(WalltimeResults {
+            instrument: Instrument {
+                type_: "walltime".to_string(),
+            },
+            creator: Creator {
+                // TODO: Stop impersonating codspeed-rust 🥸
+                name: "codspeed-rust".to_string(),
```
adriencaccia: Link an existing issue for this

crates/runner-shared/src/walltime_results.rs
adriencaccia: As seen together, you need to update the value of this variable to be seconds, not nano-seconds

GuillaumeLagrange: Indeed, I fixed the implem and imported the tests from codspeed rust

crates/runner-shared/src/walltime_results.rs
```diff
@@ -63,3 +182,95 @@ pub struct WalltimeResults {
...
+
+        let benchmark = WalltimeBenchmark::from_runtime_data(
+            NAME.to_string(),
+            URI.to_string(),
+            iters_per_round,
+            vec![42, 42 * 2, 42 * 3, 42 * 4, 42 * 5, 42 * 6],
+            None,
+        );
+
+        dbg!(vec![42, 42 * 2, 42 * 3, 42 * 4, 42 * 5, 42 * 6]);
```
adriencaccia: Remove this

crates/runner-shared/src/walltime_results.rs
```diff
@@ -44,16 +45,134 @@ pub struct WalltimeBenchmark {
...
+        };
+
+        let min_ns = times_per_iteration_per_round_ns
+            .first()
+            .copied()
+            .unwrap_or(0.0);
+        let max_ns = times_per_iteration_per_round_ns
+            .last()
+            .copied()
+            .unwrap_or(0.0);
```
adriencaccia: This is simply false, since `times_per_iteration_per_round_ns` is not sorted

adriencaccia (changes requested):

crates/runner-shared/src/walltime_results.rs
adriencaccia: Let's ensure that we have tests that actually test the whole algorithm, because it did not catch this evident issue

GuillaumeLagrange: Updated the implementation and tests after matching it against pytest-codspeed. Sorry about that!

PR #190, opened 2026-01-13
Cod 1818 do not display a table in the runner if theres only one

GuillaumeLagrange:


src/exec/mod.rs
```diff
@@ -37,16 +37,11 @@ impl ExecArgs {
     /// Merge CLI args with project config if available
     ///
     /// CLI arguments take precedence over config values.
-    pub fn merge_with_project_config(
-        mut self,
-        project_config: Option<&ProjectConfig>,
-    ) -> Self {
+    pub fn merge_with_project_config(mut self, project_config: Option<&ProjectConfig>) -> Self {
```
adriencaccia: There are changes that are only formatting of #189, they should not appear here.

PR #191, opened 2026-01-14
Cod 1815 -  Add the full command command that was run to the benchmark uri

GuillaumeLagrange:


crates/exec-harness/src/uri.rs
```diff
@@ -0,0 +1,11 @@
+pub(crate) struct NameAndUri {
+    pub(crate) name: String,
+    pub(crate) uri: String,
+}
+
+pub(crate) fn generate_name_and_uri(command: &[String]) -> NameAndUri {
```
not-matthias: If you have a very big command, this will also create a long URI. I don't recall the exact limits, but it would make sense to have an assert/anyhow::ensure here just to fail early.

src/exec/mod.rs
```diff
@@ -25,10 +25,6 @@ pub struct ExecArgs {
     #[command(flatten)]
     pub walltime_args: exec_harness::walltime::WalltimeExecutionArgs,
 
-    /// Optional benchmark name (defaults to command filename)
-    #[arg(long)]
-    pub name: Option<String>,
```
not-matthias, line 30: Do we really want to remove this? If I was a user, I might at first use the auto-generated URI but later set a custom one that is more meaningful. But we can also add it later on once people request it.

crates/exec-harness/tests/integration_test.rs
```diff
@@ -321,3 +321,126 @@ fn test_single_long_execution() -> Result<()> {
 
     Ok(())
 }
+
+/// Test that a command with shell operators (&&) works correctly when passed as a single argument
+#[test]
```
not-matthias, line 326: We also have the executor tests, which already include a lot of these edge cases: https://github.com/CodSpeedHQ/runner/blob/0cfb188332fa11041e65245f2a3cbe69ee3154ff/src/executor/tests.rs#L14-L50

We should test it there, so that we can reuse the existing test cases that ensure that the same shell scripts work across all executors.

not-matthias (changes requested):
I think we can improve the testing a bit, to avoid re-implementing the same set of shell tests in every new executor (lmk if there are any technical reasons that prevent this for the exec-harness). 

Other than that, it looks pretty good. Next review should be quick :)

crates/exec-harness/src/uri.rs
GuillaumeLagrange: We do not really have a clean handling in other integraions, I'll just truncate to a reasonable `S3 limit - 100` characters, but the real solution to this is to have a more structured metadata around the benchmarks!

crates/exec-harness/tests/integration_test.rs
GuillaumeLagrange: As discussed, I added a small subset of commands to run with the exec-harness. For now, script-like multi-commands are kinda out of scope, we'll see if we get user demand.

Kept the integration tests in the exec-harness, we'll see if we need more integration tests with pipes and whatnot form the codspeed runner

PR #193, opened 2026-01-15
Optimize perf teardown performance

GuillaumeLagrange:
Three things done here
1. Use bufwriters when writing to disk for debug_symbols
2. Serialize the whole perf-map content before saving in file for symbols
3. End fifo's loop faster after program termination

### DIFF
--- a/src/executor/shared/fifo.rs
+++ b/src/executor/shared/fifo.rs
@@ -147,7 +147,7 @@ impl RunnerFifo {
                 break;
             }
 
-            let result = tokio::time::timeout(Duration::from_secs(5), self.recv_cmd()).await;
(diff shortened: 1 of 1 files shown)


src/executor/wall_time/perf/perf_map.rs
```diff
@@ -138,22 +138,29 @@ impl ModuleSymbols {
...
-        for symbol in &self.symbols {
-            writeln!(
-                file,
-                "{:x} {:x} {}",
-                symbol.addr.wrapping_add(self.load_bias),
-                symbol.size,
-                symbol.name
-            )?;
-        }
-
```
not-matthias, line 156: Couldn't we also just use a BufWriter here? It should have the same result (minimizing the write syscalls) while being much less complex and less code.

PR #194, opened 2026-01-16
Manage memory instrument with profiling cli

GuillaumeLagrange:


crates/exec-harness/src/main.rs
```diff
@@ -1,12 +1,11 @@
-use crate::prelude::*;
```
not-matthias, line 1: Let's also define the `prelude` as a mod and use it

crates/memtrack/src/ebpf/c/memtrack.bpf.c
```diff
@@ -192,9 +192,6 @@ UPROBE_WITH_ARGS(realloc, PT_REGS_PARM2(ctx), PT_REGS_RC(ctx), EVENT_TYPE_REALLO
 /* aligned_alloc: allocates with alignment and size */
 UPROBE_WITH_ARGS(aligned_alloc, PT_REGS_PARM2(ctx), PT_REGS_RC(ctx), EVENT_TYPE_ALIGNED_ALLOC)
 
-SEC("tracepoint/syscalls/sys_enter_execve")
```
not-matthias: This is not needed anymore, I've included the same fix in the non-libc PR (now merged)

See b8535fce062a3a681672e340992a1ac531c104aa

GuillaumeLagrange: Yeah I got the conflict, this commit has been terminated :sunglasses:

PR #195, opened 2026-01-16
fix: fail if memtrack isn't supported

not-matthias:
Reworked the shared FIFO handler to first run `handle_cmd` which allows overriding the commands. Then added a check in SetVersion to check for minimum required version

### DIFF
--- a/src/executor/memory/executor.rs
+++ b/src/executor/memory/executor.rs
@@ -16,7 +16,7 @@ use async_trait::async_trait;
 use ipc_channel::ipc;
 use memtrack::MemtrackIpcClient;
 use memtrack::MemtrackIpcServer;
-use runner_shared::artifacts::{ArtifactExt, ExecutionTimestamps};
+use runner_shared::artifacts::{ArtifactExt, ExecutionTimestamps, MemtrackArtifact};
 use runner_shared::fifo::Command as FifoCommand;
 use runner_shared::fifo::IntegrationMode;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


src/executor/memory/executor.rs
```diff
@@ -154,36 +154,38 @@ impl MemoryExecutor {
...
             match cmd {
-                FifoCommand::CurrentBenchmark { pid, uri } => {
-                    debug!("Current benchmark: {pid}, {uri}");
+                FifoCommand::SetVersion(protocol_version) => {
+                    if *protocol_version < 2 {
+                        bail!(
+                            "Memory profiling requires protocol version 2 or higher, but the integration is using version {protocol_version}. \
+                            This integration doesn't support memory profiling. Please update your integration to a version that supports memory profiling.",
+                        );
+                    }
```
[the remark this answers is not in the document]
not-matthias: No, we should fallback to the default version check

src/executor/shared/fifo.rs
```diff
@@ -158,38 +162,44 @@ impl RunnerFifo {
             };
             trace!("Received command: {cmd:?}");
 
+            // Try executor-specific handler first
+            if let Some(ack) = handle_cmd(&cmd).await? {
```
GuillaumeLagrange: That's not necessarily an ack here, it's more of an "early return"

crates/memtrack/src/main.rs
```diff
@@ -177,6 +177,11 @@ fn track_command(
         }
         writer.finish()?;
 
+        if i == 0 {
```
GuillaumeLagrange: Stupid question: is this triggered if there are no allocations that are registered ?

IMO this failure would be better located in the runner teardown, WDYT ?

GuillaumeLagrange (changes requested):
tests are failing, not sure why, ping me for quick re-review

crates/memtrack/src/main.rs
not-matthias: Yeah, I agree. The test failures also seemed to be because of this check

PR #199, opened 2026-01-19
COD-2032: Allow multiple uploads in a single job

fargito:


src/run_environment/provider.rs
```diff
@@ -96,6 +97,27 @@ pub trait RunEnvironmentProvider {
...
+                );
+                match run_index_state.get_and_increment() {
+                    Ok(run_index) => Some(run_part.with_run_index(run_index)),
+                    Err(e) => {
+                        warn!("Failed to track run index: {e}. Continuing with index 0.");
+                        Some(run_part.with_run_index(0))
+                    }
+                }
+            }
+            None => None,
```
GuillaumeLagrange, line 118: You could use `.map` on `self.get_run_provider_run_part()` to avoid the match arm `None => None`

src/run/run_index_state.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// * `run_part_id` - The run part identifier (job name + matrix info)
+    pub fn new(repository_root_path: &str, run_id: &str, run_part_id: &str) -> Self {
+        // Hash the run_part_id to avoid filesystem-unsafe characters
+        // (run_part_id can contain JSON with colons, braces, quotes, etc.)
+        let run_part_id_hash = sha256::digest(run_part_id);
+        let state_file_path = PathBuf::from(repository_root_path)
+            .join(".codspeed")
+            .join("run-state")
+            .join(run_id)
+            .join(format!("{}.json", &run_part_id_hash[..16]));
```
GuillaumeLagrange, line 39: is there a particular reason we truncate this hash ?

src/run_environment/provider.rs
fargito: #200

src/run/run_index_state.rs
fargito: #200

PR #201, opened 2026-01-20
fix(runner): dont run health check in hot path

not-matthias:
- Running the perf ping everytime before we receive a command puts unnecessary load on the other end, and also has the chance of failing to receive an answer in time.
- Also added a test to ensure the FIFO is cancel-safe. This can be an issue when we read the `len` but then timeout and drop the future, since `read_exact` isn't cancel-safe. So we'll never handle this command and the integration never receives a response
- Instead of using the perf fifo to check if the process is still alive, we now check if all the sub-processes terminates. The `bash` process will still exist since it will terminate on `process.wait()` after the FIFO handler terminates (an unfortunate limitation of having a single-threaded async runtime)

### DIFF
--- a/src/executor/shared/fifo.rs
+++ b/src/executor/shared/fifo.rs
@@ -131,17 +131,22 @@ impl RunnerFifo {
         Ok(())
     }
 
-    /// Handles all incoming FIFO messages until it's closed, or until the health check closure
-    /// returns `false` or an error.
+    /// Handles all incoming FIFO messages until it's closed, or until the child process exits.
     ///
     /// The `handle_cmd` callback is invoked first for each command. If it returns `Some(response)`,
     /// that response is sent and the shared implementation is skipped. If it returns `None`,
     /// the command falls through to the shared implementation for standard handling.
+    ///
+    /// Returns execution timestamps, benchmark data, and the exit status of the child process.
     pub async fn handle_fifo_messages(
         &mut self,
-        mut health_check: impl AsyncFnMut() -> anyhow::Result<bool>,
+        child: &mut std::process::Child,
         mut handle_cmd: impl AsyncFnMut(&FifoCommand) -> anyhow::Result<Option<FifoCommand>>,
-    ) -> anyhow::Result<(ExecutionTimestamps, FifoBenchmarkData)> {
+    ) -> anyhow::Result<(
+        ExecutionTimestamps,
+        FifoBenchmarkData,
+        std::process::ExitStatus,
+    )> {
         let mut bench_order_by_timestamp = Vec::<(u64, String)>::new();
         let mut bench_pids = HashSet::<pid_t>::new();
         let mut markers = Vec::<MarkerType>::new();
@@ -241,20 +246,24 @@ impl RunnerFifo {
                 }
             }
 
-            let is_alive = health_check().await?;
-            if !is_alive {
-                debug!("Process terminated, stopping the command handler");
-                break;
+            // Check if the process has exited using try_wait (non-blocking)
+            match child.try_wait() {
+                Ok(None) => {} // Still running, continue loop
+                Ok(Some(exit_status)) => {
+                    debug!(
+                        "Process terminated with status: {exit_status}, stopping the command handler"
+                    );
+                    let marker_result =
+                        ExecutionTimestamps::new(&bench_order_by_timestamp, &markers);
+                    let fifo_data = FifoBenchmarkData {
+                        integration,
+                        bench_pids,
+                    };
+                    return Ok((marker_result, fifo_data, exit_status));
+                }
+                Err(e) => return Err(anyhow::Error::from(e)),
             }
         }
-
-        let marker_result = ExecutionTimestamps::new(&bench_order_by_timestamp, &markers);
-        let fifo_data = FifoBenchmarkData {
-            integration,
-            bench_pids,
-        };
-
-        Ok((marker_result, fifo_data))
     }
 }
 
--- a/src/executor/wall_time/perf/mod.rs
+++ b/src/executor/wall_time/perf/mod.rs
@@ -31,10 +31,7 @@ use runner_shared::metadata::PerfMetadata;
 use runner_shared::unwind_data::UnwindData;
 use std::path::Path;
 use std::path::PathBuf;
-use std::sync::Arc;
-use std::time::Duration;
 use std::{cell::OnceCell, collections::HashMap, process::ExitStatus};
-use tokio::sync::Mutex;
 
 mod jit_dump;
 mod memory_mappings;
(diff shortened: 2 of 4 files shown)


src/executor/shared/fifo.rs
```diff
@@ -145,89 +145,94 @@ impl RunnerFifo {
...
 
-            let result = tokio::time::timeout(Duration::from_secs(1), self.recv_cmd()).await;
-            let cmd = match result {
-                Ok(Ok(cmd)) => cmd,
-                Ok(Err(e)) => {
-                    warn!("Failed to parse FIFO command: {e}");
-                    break;
+        // Outer loop: continues until health check fails
+        loop {
+            // Inner loop: process commands until timeout/error
```
GuillaumeLagrange, line 166: Shouldnt this just be until timeout ?
Do we really want to continue the loops if we somehow fail to parse a command ?
(Note: your comment is in line with the implementation, I'm just ocnsidering why we wouldnt stop completely in case of error)

src/executor/shared/fifo.rs
```diff
@@ -145,89 +145,94 @@ impl RunnerFifo {
...
+            loop {
+                let result = tokio::time::timeout(Duration::from_secs(1), self.recv_cmd()).await;
+                let cmd = match result {
+                    Ok(Ok(cmd)) => cmd,
+                    Ok(Err(e)) => {
+                        warn!("Failed to parse FIFO command: {e}");
+                        break;
+                    }
+                    Err(_) => break, // Timeout
```
GuillaumeLagrange, line 176: nit: is the error necessarily a timeout ? Shouldnt we check that it is indeed a timeout and panic if something else occurs ?

src/executor/wall_time/perf/mod.rs
```diff
@@ -205,10 +202,12 @@ impl PerfRunner {
         let cmd = wrap_with_sudo(wrapped_builder)?.build();
         debug!("cmd: {cmd:?}");
 
-        let on_process_started = async |_| -> anyhow::Result<()> {
+        let on_process_started = async |pid| -> anyhow::Result<()> {
             // If we output pipedata, we do not parse the perf map during teardown yet, so we need to parse memory
             // maps as we receive the `CurrentBenchmark` fifo commands.
-            let data = Self::handle_fifo(runner_fifo, perf_fifo, self.output_pipedata).await?;
+            let data =
+                Self::handle_fifo(runner_fifo, perf_fifo, self.output_pipedata, pid as pid_t)
```
GuillaumeLagrange: can't we type pid in the closure parameter typing directly ?

src/executor/wall_time/perf/mod.rs
```diff
@@ -248,37 +247,44 @@ impl PerfRunner {
...
+        mut perf_fifo: PerfFifo,
         parse_memory_maps: bool,
+        pid: pid_t,
     ) -> anyhow::Result<BenchmarkData> {
         let mut symbols_by_pid = HashMap::<pid_t, ProcessSymbols>::new();
         let mut unwind_data_by_pid = HashMap::<pid_t, Vec<UnwindData>>::new();
 
-        let perf_fifo = Arc::new(Mutex::new(perf_fifo));
-        let mut perf_ping_timeout = 5;
+        // The runner spawns a `bash` process, which will execute perf and the benchmark. To check if it
```
GuillaumeLagrange: I think I'd rather we have an explicit thread to handle this, instead of hacking something that relies on the executor having a root bash process.

We don't necessarily need to switch to a full multi-threaded async runtime, we could explicitely spawn a thread and handle this from there. Let's create a dedicated task for this IMO.

src/executor/shared/fifo.rs
not-matthias: It's just how the timeout function works: https://docs.rs/tokio/latest/tokio/time/fn.timeout.html
> This function returns a future whose return type is [Result](https://doc.rust-lang.org/nightly/core/result/enum.Result.html)<T,[Elapsed](https://docs.rs/tokio/latest/tokio/time/error/struct.Elapsed.html)>, where T is the return type of the provided future.

We can't do pattern matching since it's a struct without any fields, but I guess we could define the type: `Err(e: Timeout) => break`

EDIT: Doesn't work, but we can type the result like this to make it more clear
```

                let result: Result<_, Elapsed> =
                    tokio::time::timeout(Duration::from_secs(1), self.recv_cmd()).await;
```

src/executor/shared/fifo.rs
not-matthias: > Do we really want to continue the loops if we somehow fail to parse a command ?

We don't have to, but I think it would be better if we panic rather than just stop the FIFO handler. Because the integration will then have UB (it'll most likely just try to send cmds which timeout, causing it to run really slowly).

src/executor/wall_time/perf/mod.rs
not-matthias: I'm not a huge fan of using an additional thread, but I found a nice workaround: We can just pass the instance to the callback, which allows us to check if the process is still alive with `try_wait`.

src/executor/shared/fifo.rs
not-matthias: I think I'll create a new ticket for this, since this could potentially break if we don't validate it properly. Don't think it makes sense to merge it in this PR, since I still have a ton of other stuff in my backlog.

src/executor/wall_time/perf/mod.rs
```diff
@@ -248,37 +246,30 @@ impl PerfRunner {
...
+        // Check if the process has exited using try_wait (non-blocking)
         let health_check = async || {
-            let perf_ping = tokio::time::timeout(Duration::from_secs(perf_ping_timeout), async {
-                perf_fifo.lock().await.ping().await
-            })
-            .await;
-            if let Ok(Err(_)) | Err(_) = perf_ping {
-                debug!("Failed to ping perf FIFO, ending perf fifo loop");
-                return Ok(false);
+            match child.try_wait() {
```
GuillaumeLagrange: The health checks are now the same. We should remove the executor specific health checks definitions and have it handled by the fifo shared implementation.

And if we need to add it later for a new instrument, we'll just make it an option with a default implementation that is this one.

src/executor/wall_time/perf/mod.rs
```diff
@@ -248,37 +246,30 @@ impl PerfRunner {
...
+                Ok(None) => Ok(true),     // Still running
+                Ok(Some(_)) => Ok(false), // Exited
```
GuillaumeLagrange: let's forward the exit code all the way

GuillaumeLagrange: We kinda do by calling the `child.wait()` afterwards, but let's take the exit code from this function, and just return it, this way reading the code it will be clearer that when we exit this function, we are sure that the process has ended.

PR #202, opened 2026-01-20
Follow frames until we reach a non inlined function in the debug info creation

GuillaumeLagrange:
We used to resolve symbols to their innermost inlined frame, which would locate some user-written code that inlines something from the stdblib inside the stdlib files.

### DIFF
--- a/src/executor/wall_time/perf/debug_info.rs
+++ b/src/executor/wall_time/perf/debug_info.rs
@@ -1,6 +1,6 @@
 use crate::executor::wall_time::perf::perf_map::ModuleSymbols;
 use crate::prelude::*;
-use addr2line::gimli;
+use addr2line::{fallible_iterator::FallibleIterator, gimli};
 use object::{Object, ObjectSection};
 use runner_shared::debug_info::{DebugInfo, ModuleDebugInfo};
 use std::path::Path;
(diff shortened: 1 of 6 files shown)


src/executor/wall_time/perf/debug_info.rs
```diff
@@ -47,13 +47,13 @@ impl ModuleDebugInfoExt for ModuleDebugInfo {
...
-                    }
-                    _ => return None,
-                };
+                // Use find_frames() instead of find_location() to handle inlined functions correctly.
+                // find_location() returns the innermost inlined location, but we want the outermost
+                // frame which is the actual function definition.
+                let frames = ctx.find_frames(symbol.addr).skip_all_loads().ok()?;
+                // Take the last frame (outermost/non-inline caller)
+                let location = frames.last().ok()??.location?;
+                let (file, line) = (location.file?.to_string(), location.line);
```
art049, line 67: The comments are not really clear. IIUC, you go as deep as possible in the callstack because inlined functions (which are not frames) are not there.
Talking about the outermost caller is a bit confusing.
let's say I have `foo` -> `bar` -> `baz(inlined)`
When I'm in baz inlined we want to return bar but the "outermost caller" as you mentioned, is technically `foo`
So it should be the innermost non inline function maybe? or more simply innermost frame

art049 (approved):
lgtm except the clarity in the comment

src/executor/wall_time/perf/debug_info.rs
GuillaumeLagrange: Updated it to

```
                // Use find_frames() instead of find_location() to handle inlined functions correctly.
                //
                // If we have foo -> bar -> baz(inlined) -> stdfunc(inlined)
                // where the whole body of bar is the inlined baz, which itself is just inlined stdfunc.
                //
                // Using find_location() on the `bar` symbol address would return the location of
                // `stdfunc`, while using find_frames() an iterator that yields the frames in
                // order:
                // 1. stdfunc (inlined)
                // 2. baz (inlined)
                // 3. bar
                //
                // And stops until a non inlined function is reached.
                // We can then take the last frame to get the correct location.
```

GuillaumeLagrange: I tested it, empirically, which is good enough for me now, if we re-assess this we may want to look into DWARF specs, but this should hold

PR #209, opened 2026-01-22
Manage valgrind instrumentation with profiling cli

GuillaumeLagrange:
- Add a small c library that enables and disables valgrind instrumentation at dynamic link through `constructor` and `destructor` attributes (which are supported by both `gcc` and `clang`)
- Embed the library into the exec-harness binary
- Load this library using `LD_PRELOAD`, after writing it to disk at runtime

### DIFF
--- a/crates/exec-harness/Cargo.toml
+++ b/crates/exec-harness/Cargo.toml
@@ -27,4 +27,4 @@ cargo_metadata = "0.19"
 cc = "1"
 
 [package.metadata.dist]
-targets = ["aarch64-unknown-linux-musl", "x86_64-unknown-linux-musl"]
+targets = ["aarch64-unknown-linux-gnu", "x86_64-unknown-linux-gnu"]


crates/exec-harness/src/analysis.rs
```diff
@@ -27,3 +32,72 @@ pub fn perform(commands: Vec<BenchmarkCommand>) -> Result<()> {
...
+}
+
+/// Returns the path to the preload library, extracting it to a temp file if needed.
+fn get_preload_lib_path() -> Result<&'static std::path::Path> {
+    if let Some(path) = PRELOAD_LIB_PATH.get() {
+        return Ok(path);
+    }
+
+    let path = extract_preload_lib()?;
+    Ok(PRELOAD_LIB_PATH.get_or_init(|| path))
```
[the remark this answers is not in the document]
GuillaumeLagrange: `get_or_try_init` is nightly bro

crates/exec-harness/preload/codspeed_preload.c
```diff
@@ -0,0 +1,101 @@
...
+#warning "RUNNING_ON_VALGRIND not defined, headers may be missing"
+#define RUNNING_ON_VALGRIND 0
+#endif
+
+static const char *LOCK_ENV = "CODSPEED_PRELOAD_LOCK";
+
+// These constants are defined by the build script (build.rs) via -D flags
+#ifndef CODSPEED_URI_ENV
+#error "CODSPEED_URI_ENV must be defined by the build system"
+#endif
```
not-matthias, line 30: Why are we passing this via the build script? can't this be static like the lock env above?

crates/exec-harness/preload/codspeed_preload.c
```diff
@@ -0,0 +1,101 @@
...
+    instrument_hooks_deinit(g_hooks);
+    g_hooks = NULL;
+    return;
+  }
+}
+
+__attribute__((destructor)) static void codspeed_preload_fini(void) {
+  // If the process is not the owner of the lock, this means g_hooks was not
+  // initialized
+  if (!g_hooks) {
```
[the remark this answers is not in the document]
not-matthias: Couldn't we remove the `g_is_owner`? If `g_hooks` is NULL, then we aren't the owner

crates/exec-harness/build.rs
```diff
@@ -0,0 +1,196 @@
...
+        .arg("-lpthread");
+
+    let status = link_cmd.status().expect("Failed to run linker");
+    if !status.success() {
+        panic!("Failed to link libcodspeed_preload.so");
+    }
+}
+
+/// Find the instrument-hooks directory from the codspeed crate using cargo_metadata
+fn find_codspeed_instrument_hooks_dir() -> PathBuf {
```
not-matthias, line 132: Just double checking: This is using the artifacts from the _current_ project, not the global `~/.cargo/registry/`, right?

crates/exec-harness/build.rs
```diff
@@ -0,0 +1,196 @@
...
+    let instrument_hooks_dir = codspeed_dir.join("instrument-hooks");
+
+    if !instrument_hooks_dir.exists() {
+        panic!("instrument-hooks directory not found at {instrument_hooks_dir}");
+    }
+
+    instrument_hooks_dir.into_std_path_buf()
+}
+
+impl PreloadBuildPaths {
```
not-matthias, line 180: nit: this should be below the struct definition

crates/exec-harness/build.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// This should match the version of the `codspeed` crate dependency.
+    integration_version: &'static str,
+}
+
+// TODO(COD-1736): Stop impersonating codspeed-rust 🥸
+const PRELOAD_CONSTANTS: PreloadConstants = PreloadConstants {
+    uri_env: "CODSPEED_BENCH_URI",
+    integration_name: "codspeed-rust",
+    integration_version: "4.2.0",
+};
```
not-matthias: nit: I'm not sure how I feel about having this global here 🤔 The struct is fine because it makes it nicer, but then we use the consts in `main` but pass it as an arg to `build_shared_library`?

Are we planning on having multiple preload files? If no, then we could add a `Default` impl or `::new()` func that just gives us the current constants and instantiate it in `main()`.

crates/exec-harness/build.rs
```diff
@@ -0,0 +1,196 @@
...
+    let integration_name_val = format!("\"{}\"", constants.integration_name);
+    let integration_version_val = format!("\"{}\"", constants.integration_version);
+
+    let mut build = cc::Build::new();
+    build
+        .file(&paths.preload_c)
+        .file(&paths.core_c)
+        .include(&paths.includes_dir)
+        .pic(true)
+        .opt_level(2)
```
not-matthias: Why not -O3?

crates/exec-harness/build.rs
```diff
@@ -0,0 +1,196 @@
...
+        .cargo_metadata(false)
```
not-matthias, line 94: Had to lookup why we need this, maybe add a comment for the next person reading this: 
> Define whether metadata should be emitted for cargo allowing it to automatically link the binary. Defaults to true.

crates/exec-harness/preload/codspeed_preload.c
GuillaumeLagrange: this is to make sure that the name of the env var cannot be desync between the built C code, and the exec-harness.

crates/exec-harness/build.rs
GuillaumeLagrange: As discussed, this will resolve dynamicly to the correct `codspeed` path, to make sure that we have one and only one instrument-hooks implementation used

PR #212, opened 2026-01-23
feat(memtrack): add C++ and Rust integration tests

not-matthias:
- Add C++ and Rust tests using common allocator functions
- Also fixes a few issues discovered by the tests:
  - Adds support for previously missing `posix_memalign` and `libstdc++` (libc but for C++)
  - Detect non-default allocators first for binaries
  - Attach to different jemalloc prefixed functions

### DIFF
--- a/crates/memtrack/src/ebpf/memtrack.rs
+++ b/crates/memtrack/src/ebpf/memtrack.rs
@@ -262,40 +262,6 @@ impl MemtrackBpf {
     // Attach methods grouped by allocator
     // =========================================================================
 
-    /// Attach standard library allocation probes (libc-style: malloc, free, calloc, etc.)
-    /// This works for libc and allocators that export standard symbol names.
-    /// For non-libc allocators, standard names are optional - just try them silently.
-    pub fn attach_libc_probes(&mut self, lib_path: &Path) -> Result<()> {
-        self.try_attach_malloc(lib_path, "malloc");
-        self.try_attach_calloc(lib_path, "calloc");
-        self.try_attach_realloc(lib_path, "realloc");
-        self.try_attach_free(lib_path, "free");
-        self.try_attach_aligned_alloc(lib_path, "aligned_alloc");
-        self.try_attach_memalign(lib_path, "posix_memalign");
-        self.try_attach_memalign(lib_path, "memalign");
-        Ok(())
-    }
-
-    /// Attach C++ operator new/delete probes.
-    /// These are mangled C++ symbols that wrap the underlying allocator.
-    /// C++ operators have identical signatures to malloc/free, so we reuse those handlers.
-    pub fn attach_libcpp_probes(&mut self, lib_path: &Path) -> Result<()> {
-        self.try_attach_malloc(lib_path, "_Znwm"); // operator new(size_t)
-        self.try_attach_malloc(lib_path, "_Znam"); // operator new[](size_t)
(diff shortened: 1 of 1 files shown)


crates/memtrack/tests/shared.rs
```diff
@@ -0,0 +1,93 @@
...
+use memtrack::prelude::*;
+use memtrack::{AllocatorLib, EventType, MemtrackEventExt, Tracker};
+use runner_shared::artifacts::{MemtrackEvent as Event, MemtrackEventKind};
+use std::path::Path;
+use std::process::Command;
+use std::time::Duration;
+
+type TrackResult = anyhow::Result<(Vec<Event>, std::thread::JoinHandle<()>)>;
+
+macro_rules! assert_events_with_marker {
```
GuillaumeLagrange: Can we add an example usage as docstring here

crates/memtrack/tests/c_tests.rs
```diff
@@ -1,38 +1,13 @@
-use libbpf_rs::ErrorExt;
-use memtrack::{EventType, MemtrackEventExt, Tracker};
+mod shared;
```
GuillaumeLagrange, line 2: couldnt we get rid of all the checking logic and add snapshots here as well ?

crates/memtrack/src/ebpf/memtrack.rs
```diff
@@ -340,7 +340,27 @@ impl MemtrackBpf {
         // - rust_dealloc: _rjem_sdallocx
         // - rust_realloc: _rjem_realloc / _rjem_rallocx
 
-        // Prefixed standard API
+        // je_*_default API (C++ with static linking)
+        self.try_attach_malloc(lib_path, "je_malloc_default");
```
GuillaumeLagrange: I can see this becoming quite an annoying thing to maintain over time.

Could we have some sort of structured consts for each allocator that would just be

```rust
struct AllocatorSymbols {
   pub malloc_symbols = [...];
   pub ...
}
```

then each prob attach function would just loop over each symbol and call the proper functoin

Feel free to disregard if you dont see the point

crates/memtrack/tests/c_tests.rs
not-matthias: Good catch, that makes it much simpler :+1:

crates/memtrack/src/ebpf/memtrack.rs
not-matthias: Yeah, 100% agree. I think having a struct with malloc symbols is a bit overkill, since there aren't _that_ many. But I've added a new function `attach_standard_probes` which takes a prefix and postfix array. This way we can handle all the variants with their own naming patterns: 
```rust
        let prefixes = ["je_", "_rjem_"];
        let suffixes = ["", "_default"];

        self.attach_standard_probes(lib_path, &prefixes, &suffixes)?;

        // Non-standard API that has an additional flag parameter
        // See: https://jemalloc.net/jemalloc.3.html
        for prefix in prefixes {
            for suffix in suffixes {
                self.try_attach_malloc(lib_path, &format!("{prefix}mallocx{suffix}"));
                self.try_attach_realloc(lib_path, &format!("{prefix}rallocx{suffix}"));
                self.try_attach_free(lib_path, &format!("{prefix}dallocx{suffix}"));
                self.try_attach_free(lib_path, &format!("{prefix}sdallocx{suffix}"));
            }
        }
```
For the non-default API, we'll still have to manually match it and ensure we use the right probe.

PR #213, opened 2026-01-23
feat(runner): add go-runner-version CLI argument

not-matthias:


src/executor/config.rs
```diff
@@ -121,6 +125,7 @@ impl TryFrom<RunArgs> for Config {
             skip_run: args.shared.skip_run,
             skip_setup: args.shared.skip_setup,
             allow_empty: args.shared.allow_empty,
+            go_runner_version: args.shared.go_runner_version.clone(),
```
GuillaumeLagrange: there should be no need to clone here we have ownership

PR #215, opened 2026-01-23
Create codspeed yaml schema

GuillaumeLagrange:


Cargo.toml
```diff
@@ -4,11 +4,15 @@ version = "4.9.0"
...
 publish = false
+default-run = "codspeed"
 
 [[bin]]
 name = "codspeed"
 path = "src/main.rs"
 
+[[bin]]
+name = "generate-config-schema"
+path = "src/bin/generate_config_schema.rs"
```
art049: normally this is unnecessary, everything under src/bin is consider a bin target by default
https://doc.rust-lang.org/cargo/guide/project-layout.html

Cargo.toml
```diff
@@ -4,11 +4,15 @@ version = "4.9.0"
 edition = "2024"
 repository = "https://github.com/CodSpeedHQ/codspeed"
 publish = false
+default-run = "codspeed"
```
art049: as well normally the src/main.rs is probably considered the default run

Cargo.toml
GuillaumeLagrange: indeed, removed it

Cargo.toml
GuillaumeLagrange: It's still necessary, I'm not sure whether it's because of the weird structure (root cargo toml being both the workspace and a crate), or because of the name.

I removed the explicit bin declaration, but kept this

PR #217, opened 2026-01-25
feat: improve the result table display

art049:
image


src/run/helpers/benchmark_display.rs
```diff
@@ -0,0 +1,386 @@
...
+
+#[derive(Tabled)]
+struct WalltimeRow {
+    #[tabled(rename = "Benchmark")]
+    name: String,
+    #[tabled(rename = "Time (best)")]
+    time_best: String,
+    #[tabled(rename = "Iterations")]
+    iterations: String,
+    #[tabled(rename = "Rel. StdDev")]
```
GuillaumeLagrange: Is `Rel. StdDev` explicit enough ? Does not cost much to have `Relative StdDev` here

art049: It creates a really wide row

art049: Actually maybe we could just put StdDev and put it in percent and it would be clear it's relative, WDYT?

GuillaumeLagrange: OKFM, % implies relative clearly enough IMO

art049: 👍

art049: Will add it later along with more data for the simulation instrument 😎

PR #219, opened 2026-01-26
fix(memtrack): track previous addr of realloc calls

not-matthias:


crates/memtrack/src/ebpf/c/memtrack.bpf.c
```diff
@@ -123,77 +123,187 @@ static __always_inline __u64* take_param(void* map) {
...
-static __always_inline int submit_event(__u64 addr, __u64 size, __u8 event_type) {
-    __u64 tid = bpf_get_current_pid_tgid();
-    __u32 pid = tid >> 32;
-
-    if (!is_tracked(pid) || !is_enabled()) {
-        return 0;
+/* Macro to handle common event submission boilerplate
+ * Usage: SUBMIT_EVENT(event_type, { e->data.foo = bar; })
+ */
+#define SUBMIT_EVENT(evt_type, fill_data)                              \
```
GuillaumeLagrange, line 132: side note: I think your fomratter is a bit MIA here, with how the `\` line on the right looks. De we have automated format checking in these files ? Could be nice to have

not-matthias: We didn't have a formatter for this file yet, I added clang-format

PR #220, opened 2026-01-26
fix: rename `targets` to `benchmarks` in the project_config to match spec

GuillaumeLagrange:
Keeping `targets` alias to avoid breaking changes


src/project_config/interfaces.rs
```diff
@@ -11,7 +11,8 @@ pub struct ProjectConfig {
     /// Default options to apply to all benchmark runs
     pub options: Option<ProjectOptions>,
     /// List of benchmark targets to execute
-    pub targets: Option<Vec<Target>>,
+    #[serde(alias = "target")]
```
art049: as discussed let's remove this

PR #221, opened 2026-01-27
Implement codspeed use mode to define mode for shell session

GuillaumeLagrange:
Also finished the refactor that was started when first implementing exec-harness to remove a lot of the logc that was nested in the `run` module

### DIFF
--- a/src/runner_mode/shell_session.rs
+++ b/src/runner_mode/shell_session.rs
@@ -1,0 +1,77 @@
+use super::RunnerMode;
+use crate::prelude::*;
+use libc::pid_t;
+use std::path::Path;
+use std::path::PathBuf;
+use std::sync::OnceLock;
+use sysinfo::Pid;
+use sysinfo::ProcessRefreshKind;
+use sysinfo::RefreshKind;
+use sysinfo::System;
+
+static SYSTEM: OnceLock<System> = OnceLock::new();
+
+/// Get the root directory where the use mode is stored
+/// If available, uses `$XDG_RUNTIME_DIR/codspeed_use_mode`
+/// Otherwise, falls back to `std::env::temp_dir()/codspeed_use_mode`
+fn get_use_mode_root_dir() -> PathBuf {
+    let base_dir = if let Some(xdg_runtime_dir) = std::env::var_os("XDG_RUNTIME_DIR") {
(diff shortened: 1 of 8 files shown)


src/runner_mode/shell_session.rs
```diff
@@ -0,0 +1,77 @@
...
+    let current_pid = Pid::from_u32(pid as u32);
+
+    s.process(current_pid)
+        .and_then(|p| p.parent())
+        .map(|pid| pid.as_u32() as pid_t)
+}
+
+fn get_mode_file_path(base_dir: &Path, pid: pid_t) -> PathBuf {
+    base_dir.join(pid.to_string())
+}
```
not-matthias, line 43: Why are we getting the file path with the pid here? Isn't `codspeed use {mode}` going to be executed separately? So we won't have any way of knowing which pid it had. 

Also, I'm sure about writing the mode to disk. My first impression of the `codspeed use` cmd would be that it's temporary (I guess that's why we have the pid?) and that it will be reset in a new shell (just like `nix-shell -p` for example).

GuillaumeLagrange: The pid used to register is the parent pid
The pid used to retrieve is the first pid in the process ancestry that has an entry in the mode_file_paths

As to why we wrtie to disk, unfortunately as you pointed out, `codspeed use xxx` is going to be a dedicated pid, which cannot export an env var to the caller.

The only way to remediate, is to 
1. have a wrapper function in the user shell that just exports an env variable
2. have codspeed use spawn a child shell process
3. write somewhere deterministic to a file

On linux, we write in `XDG_RUNTIME_DIR`, which is (source arch wiki)

```
    XDG_RUNTIME_DIR
        Used for non-essential, user-specific data files such as sockets, named pipes, etc.
        Not required to have a default value; warnings should be issued if not set or equivalents provided.
        Must be owned by the user with an access mode of 0700.
        Filesystem fully featured by standards of OS.
        Must be on the local filesystem.
        May be subject to periodic cleanup.
        Modified every 6 hours or set sticky bit if persistence is desired.
        Can only exist for the duration of the user's login.
        Should not store large files as it may be mounted as a tmpfs.
        pam_systemd sets this to /run/user/$UID.
```

On other platforms, we write to `$TMPDIR`, which is similar but less ad-hoc

PR #225, opened 2026-01-28
feat(runner): support pytest-codspeed memory profiling

not-matthias:
see https://github.com/CodSpeedHQ/pytest-codspeed/pull/103

### DIFF
--- a/src/executor/memory/executor.rs
+++ b/src/executor/memory/executor.rs
@@ -177,6 +177,7 @@ impl MemoryExecutor {
                     let min_version = match name.as_str() {
                         "codspeed-rust" => Version::new(4, 2, 0),
                         "codspeed-cpp" => Version::new(2, 1, 0),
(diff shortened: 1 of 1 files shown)


src/executor/memory/executor.rs
```diff
@@ -177,6 +177,7 @@ impl MemoryExecutor {
                     let min_version = match name.as_str() {
                         "codspeed-rust" => Version::new(4, 2, 0),
                         "codspeed-cpp" => Version::new(2, 1, 0),
+                        "pytest-codspeed" => Version::new(4, 2, 0), // FIXME: Change to 4.3.0
```
GuillaumeLagrange: change the version and merge this already IMO, and until we release we can either change the version in pytest or in the runner we're building

GuillaumeLagrange: do as you see fit though, this change is minor enough

PR #228, opened 2026-01-29
feat(runner): support codspeed-node memory profiling

not-matthias:


GuillaumeLagrange (approved):
Remove the comment and we can merge

PR #230, opened 2026-02-02
fix(runner): check benchmark URIs rather than memtrack events

not-matthias:
Memory profiling can be empty, if the benchmark has no allocations. We shouldn't fail in those cases.


GuillaumeLagrange (approved):
lgtm, dont forget to test both success and failure to make sure we did not invert a boolean this time 🙃

PR #231, opened 2026-02-02
Import instrument-hooks directly to set exec-harness metadata

GuillaumeLagrange:


crates/exec-harness/build.rs
```diff
@@ -185,13 +149,13 @@ impl PreloadBuildPaths {
     fn check_sources_exist(&self) {
         if !self.core_c.exists() {
             panic!(
-                "core.c not found at {}. Make sure the codspeed crate is available.",
+                "core.c not found at {}. Make sure the instrument-hooks-bindings crate is available.",
```
not-matthias, line 152: Not a big deal but the message here is a bit misleading. This panic will only happen when the Git submodule isn't available, not the crate

src/executor/memory/executor.rs
```diff
@@ -121,28 +121,27 @@ impl Executor for MemoryExecutor {
     }
 
     async fn teardown(&self, execution_context: &ExecutionContext) -> Result<()> {
-        let files: Vec<_> = std::fs::read_dir(execution_context.profile_folder.join("results"))?
+        let results_dir = execution_context.profile_folder.join("results");
```
not-matthias, line 124: Was this meant to be included? I guess you cherry-picked the commit from the other PR?

GuillaumeLagrange: Yep exactly, I needed to test a 0 alloc C++ bench at some point to make sure

GuillaumeLagrange:
Merging alongside #235

PR #235, opened 2026-02-05
Support python in profiling cli

GuillaumeLagrange:


src/executor/helpers/harvest_perf_maps_for_pids.rs
```diff
@@ -1,14 +1,13 @@
...
-    profile_folder: &Path,
-    pids: &HashSet<libc::pid_t>,
-) -> Result<()> {
+pub async fn harvest_perf_maps_for_pids<I>(profile_folder: &Path, pids: I) -> Result<()>
+where
+    I: IntoIterator<Item = libc::pid_t>,
+{
     let perf_maps = pids
-        .iter()
+        .into_iter()
```
not-matthias: Dont forget to dedup the pids here, that's why we used `HashSet`

src/executor/wall_time/perf/jit_dump.rs
```diff
@@ -102,7 +99,10 @@ impl JitDump {
 }
 
 /// Converts all the `jit-<pid>.dump` into unwind data and copies it to the profile folder.
-pub async fn harvest_perf_jit_for_pids(profile_folder: &Path, pids: &HashSet<i32>) -> Result<()> {
+pub async fn harvest_perf_jit_for_pids<I>(profile_folder: &Path, pids: I) -> Result<()>
+where
+    I: IntoIterator<Item = libc::pid_t>,
+{
     for pid in pids {
```
not-matthias, line 109: same comment here, make sure we're iterating only over the dedup/unique pids. Otherwise we can easily run into issues where we duplicate work and/or overwrite files

src/executor/helpers/harvest_perf_maps_for_pids.rs
```diff
@@ -1,14 +1,13 @@
 use crate::prelude::*;
-use std::collections::HashSet;
 use std::fs;
```
not-matthias: Since this function is async, and the `BenchmarkData::save_to` function is now also async we could switch to `tokio::fs` instead of `std::fs` here.

crates/exec-harness/src/analysis/mod.rs
```diff
@@ -54,6 +54,7 @@ pub fn perform_with_valgrind(commands: Vec<BenchmarkCommand>) -> Result<()> {
         cmd.args(&benchmark_cmd.command[1..]);
         // Use LD_PRELOAD to inject instrumentation into the child process
         cmd.env("LD_PRELOAD", preload_lib_path);
+        cmd.env("PYTHONPERFSUPPORT", "1");
```
not-matthias, line 59: nit: maybe add a quick comment that we need this so that python generates perf maps (technically it's in the commit message, but comments make it a bit more obvious)

src/executor/wall_time/perf/mod.rs
```diff
@@ -105,9 +105,9 @@ impl PerfRunner {
             || config.command.contains("uv")
             || config.command.contains("python")
         {
-            // Max supported stack size is 64KiB, but this will increase the file size by a lot. In
-            // order to allow uploads and maintain accuracy, we limit this to 8KiB.
-            (UnwindingMode::Dwarf, Some(8 * 1024))
+            // Note that the higher the stack size, the larger the file, although it is mitigated
+            // by zstd compression
+            (UnwindingMode::Dwarf, Some(32 * 1024))
```
not-matthias: Can you test this on pytest-codspeed and check the final report sizes? Last time when we tried it, we had some issues there. 

It would also be useful to know when we could run into those limits since we're saving 4x more data now.

crates/exec-harness/src/analysis/mod.rs
```diff
@@ -66,3 +71,35 @@ pub fn perform_with_valgrind(commands: Vec<BenchmarkCommand>) -> Result<()> {
 
     Ok(())
 }
+
+fn bail_if_command_spawned_subprocesses_under_valgrind(pid: u32) -> Result<()> {
```
not-matthias, line 89: can we have a comment here that explains a bit why and how this works (basically what we discussed in the DMs); don't spend too much time on it, just want to make sure we have an explanation when we come back to this code

src/executor/wall_time/perf/jit_dump.rs
GuillaumeLagrange: Moved them back to HashSet, didnt make much sense to do it like I ded

src/executor/wall_time/perf/mod.rs
GuillaumeLagrange: I've tested it on pytest-codspeed. There is an impact, but with compression + file streaming we'll be fine.

PR #236, opened 2026-02-07
feat: add tracegrind support as an alternative simulation tool

art049:


src/executor/valgrind/measure.rs
```diff
@@ -19,35 +19,41 @@ use std::path::Path;
...
+
+    match tool {
+        SimulationTool::Callgrind => {
+            args.push("--tool=callgrind".to_string());
+            args.push("--compress-strings=no".to_string());
+            args.push("--combine-dumps=yes".to_string());
+            args.push("--dump-line=no".to_string());
+        }
+        SimulationTool::Tracegrind => {
+            args.push("--tool=tracegrind".to_string());
```
GuillaumeLagrange, line 47: as discussed druing daily, let's print a warning saying that this is still in development and **will** crash for "normal" users.

PR #237, opened 2026-02-09
fix: memory profiling fails when using sub-folders

not-matthias:
- **fix(memory): handle working-directory option**
- **fix(memtrack): search build directories in sub-folders**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
 
       # Since we ran the tests with sudo, the build artifacts will have root ownership
       - name: Clean up
-        run: sudo chown -R $USER:$USER .
+        run: sudo chown -R $USER:$USER . ~/.cargo
 
   benchmarks:
     runs-on: ubuntu-latest


crates/memtrack/src/allocators/static_linked.rs
```diff
@@ -4,35 +4,65 @@ use std::path::{Path, PathBuf};
...
+    }
 
-        // Check for CMake build directory
-        let cmake_build = current_dir.join("build");
-        if cmake_build.is_dir() {
-            dirs.push(cmake_build);
+    // Walk downward from current directory
+    let mut stack = vec![(current_dir, 0)];
+    while let Some((dir, depth)) = stack.pop() {
+        const MAX_SEARCH_DEPTH: usize = 4;
```
GuillaumeLagrange: Why an arbitrary four here ?

not-matthias: It's meant to be a tradeoff between speed and accuracy. But I guess we could remove it since source dirs should be not _that_ big. I originally tested it with `/nix/store` (which has 142140 entries in the top folder) but I guess it's not comparable for dev dirs. I'll check against the monorepo how fast it would be with the depth removed

GuillaumeLagrange: At lest we should have a comment explaining how we chose 4 IMO

not-matthias: Okay, just tested without the depth limit and it's already super fast so we shouldn't need it: 
- Monorepo: 300ms
- LLVM: 4sec
- Linux: 4sec

So even on huge code bases it's quite fast.

PR #240, opened 2026-02-11
Deduplicate walltime perf symbols, unwind data and debug info across pids

GuillaumeLagrange:
# Merge/deployment
1. Merge #246 (on top of which this PR is based)
2. Merge the backend PR refering to the changed perf metadata changes
3. Merge this pr and release runner once the changes in the backend are in prod

# Content

Since the real deduplicator source of truth is actually the path of the elf file on disk, I settled on 

```
{path_index}__{file_name}.unwind_data
{path_index}__{file_name}.symbols.map
```

it allows for the filename to be determined by the actual path, without having to either nest all the files, or have to sanitize paths to file name and end up in awkward file-name length limitations

Here's what it looks like on a 1000 iterations report of one simple program
```
$ ls
0__exec-harness.symbols.map          3__ld-linux-x86-64.so.2.unwind_data  7__nix-ld.symbols.map         perf.metadata
0__exec-harness.unwind_data          4__ld-linux-x86-64.so.2.symbols.map  7__nix-ld.unwind_data         perf.pipedata
1__libm.so.6.symbols.map             4__ld-linux-x86-64.so.2.unwind_data  8__libc.so.6.symbols.map      results
1__libm.so.6.unwind_data             5__libgcc_s.so.1.symbols.map         8__libc.so.6.unwind_data      runner.log
2__graph-benchmark.symbols.map       5__libgcc_s.so.1.unwind_data         9__libgcc_s.so.1.symbols.map
2__graph-benchmark.unwind_data       6__libc.so.6.symbols.map             9__libgcc_s.so.1.unwind_data
3__ld-linux-x86-64.so.2.symbols.map  6__libc.so.6.unwind_data             ExecutionTimestamps.msgpack


```

### DIFF
--- a/src/executor/wall_time/perf/mod.rs
+++ b/src/executor/wall_time/perf/mod.rs
@@ -321,17 +321,16 @@ impl BenchmarkData {
                 .clone()
                 .ok_or(BenchmarkDataSaveError::MissingIntegration)?,
             uri_by_ts: self.marker_result.uri_by_ts.clone(),
-            ignored_modules: artifacts.ignored_modules,
+            ignored_modules_by_pid: artifacts.ignored_modules_by_pid,
             markers: self.marker_result.markers.clone(),
             debug_info: artifacts.debug_info,
             mapped_process_debug_info_by_pid: artifacts.mapped_process_debug_info_by_pid,
             mapped_process_unwind_data_by_pid: artifacts.mapped_process_unwind_data_by_pid,
             mapped_process_module_symbols: artifacts.symbol_pid_mappings_by_pid,
             path_key_to_path: artifacts.key_to_path,
-            // Not yet implemented
-            ignored_modules_by_pid: Default::default(),
             // Deprecated fields below are no longer used
             debug_info_by_pid: Default::default(),
+            ignored_modules: Default::default(),
         };
         metadata.save_to(&path).unwrap();
 
--- a/src/executor/wall_time/perf/save_artifacts.rs
+++ b/src/executor/wall_time/perf/save_artifacts.rs
@@ -16,7 +16,7 @@ pub struct SavedArtifacts {
     pub debug_info: HashMap<String, ModuleDebugInfo>,
     pub mapped_process_debug_info_by_pid: HashMap<pid_t, Vec<MappedProcessDebugInfo>>,
     pub mapped_process_unwind_data_by_pid: HashMap<pid_t, Vec<MappedProcessUnwindData>>,
-    pub ignored_modules: Vec<(String, u64, u64)>,
+    pub ignored_modules_by_pid: HashMap<pid_t, Vec<(String, u64, u64)>>,
     pub key_to_path: HashMap<String, PathBuf>,
 }
 
@@ -43,7 +43,7 @@ pub fn save_artifacts(
         &mut path_to_key,
     );
 
-    let ignored_modules = collect_ignored_modules(mounted_modules_by_path);
+    let ignored_modules_by_pid = collect_ignored_modules(loaded_modules_by_path);
 
     let key_to_path = path_to_key
         .into_iter()
@@ -55,7 +55,7 @@ pub fn save_artifacts(
         debug_info,
         mapped_process_debug_info_by_pid,
         mapped_process_unwind_data_by_pid,
-        ignored_modules,
+        ignored_modules_by_pid,
         key_to_path,
     }
 }
@@ -247,10 +247,11 @@ fn save_unwind_data(
 }
 
 /// Collect ignored modules by finding known-ignored and python modules in the mounted modules.
+/// Returns per-pid entries with runtime address ranges derived from symbol bounds + load bias.
 fn collect_ignored_modules(
-    mounted_modules_by_path: &HashMap<PathBuf, MountedModule>,
-) -> Vec<(String, u64, u64)> {
-    let mut to_ignore = vec![];
+    loaded_modules_by_path: &HashMap<PathBuf, LoadedModule>,
+) -> HashMap<pid_t, Vec<(String, u64, u64)>> {
+    let mut by_pid: HashMap<pid_t, Vec<(String, u64, u64)>> = HashMap::new();
 
     let ignore_paths = get_objects_path_to_ignore();
 
@@ -269,16 +270,25 @@ fn collect_ignored_modules(
             continue;
         }
 
-        for pm in m.process_mounted_module.values() {
-            if let Some(ref pud) = pm.process_unwind_data {
-                to_ignore.push((
+        let addr_bounds = loaded_module
+            .module_symbols
+            .as_ref()
+            .and_then(|ms| ms.addr_bounds());
+
+        let Some((elf_start, elf_end)) = addr_bounds else {
+            continue;
+        };
+
+        for (&pid, pm) in &loaded_module.process_loaded_modules {
+            if let Some(load_bias) = pm.symbols_load_bias {
+                by_pid.entry(pid).or_default().push((
                     path_str.to_string(),
-                    pud.avma_range.start,
-                    pud.avma_range.end,
+                    elf_start + load_bias,
+                    elf_end + load_bias,
                 ));
             }
         }
     }
 
-    to_ignore
+    by_pid
 }
--- a/src/executor/wall_time/perf/debug_info.rs
+++ b/src/executor/wall_time/perf/debug_info.rs
@@ -5,7 +5,9 @@ use addr2line::{fallible_iterator::FallibleIterator, gimli};
 use object::{Object, ObjectSection};
 use rayon::prelude::*;
 use runner_shared::debug_info::{DebugInfo, ModuleDebugInfo};
+use std::collections::HashMap;
 use std::path::Path;
+use std::path::PathBuf;
 
 type EndianRcSlice = gimli::EndianRcSlice<gimli::RunTimeEndian>;
 
--- a/src/executor/wall_time/perf/module_symbols.rs
+++ b/src/executor/wall_time/perf/module_symbols.rs
@@ -43,6 +43,18 @@ impl ModuleSymbols {
         &self.symbols
     }
 
+    /// Returns `(min_addr, max_addr_end)` covering all symbols, or `None` if empty.
+    pub fn addr_bounds(&self) -> Option<(u64, u64)> {
+        let first = self.symbols.first()?;
+        Some(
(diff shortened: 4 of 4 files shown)


crates/runner-shared/src/metadata.rs
```diff
@@ -25,9 +28,21 @@ pub struct PerfMetadata {
...
-    /// Debug info for all modules across all processes, mapping PID to module debug info
-    #[serde(default, skip_serializing_if = "std::collections::HashMap::is_empty")]
-    pub debug_info_by_pid: std::collections::HashMap<i32, Vec<ModuleDebugInfo>>,
+    /// Deduplicated debug info entries, keyed by semantic key
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub debug_info: HashMap<String, ModuleDebugInfo>,
+
+    /// Per-pid debug info references, mapping PID to list of debug info index + load bias
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub debug_info_pid_mappings_by_pid: HashMap<i32, Vec<DebugInfoPidMapping>>,
```
not-matthias: let's use libc::pid_t here

not-matthias: Le'ts also do it for the two other fields below

src/executor/wall_time/perf/snapshots/codspeed_runner__executor__wall_time__perf__unwind_data__tests__cpp_unwind_data.snap
```diff
@@ -1,17 +1,22 @@
...
+            eh_frame_hdr_svma: 4577bc..458b30,
+            eh_frame_hdr_hash: 4b4eac90f7f5e60d,
+            eh_frame_hash: 233bdd4ae9fe4ba4,
+            eh_frame_svma: 451098..4577bc,
+        },
+        UnwindDataPidMappingWithFullPath {
+            path: "testdata/perf_map/cpp_my_benchmark.bin",
+            timestamp: None,
+            avma_range: 4194304..4558848,
+            base_avma: 4194304,
```
not-matthias: nit: can we have a custom Debug impl to always have hex values here?

src/executor/wall_time/perf/parse_perf_file.rs
```diff
@@ -10,10 +10,34 @@ use runner_shared::unwind_data::UnwindData;
...
+pub struct UnwindDataPidMappingWithFullPath {
+    pub path: String,
+    pub timestamp: Option<u64>,
+    pub avma_range: std::ops::Range<u64>,
+    pub base_avma: u64,
+}
+
+/// Per-pid mounting info for symbols, referencing the deduplicated symbols by path.
+#[derive(Debug)]
+pub struct SymbolPidMappingWithFullPath {
```
not-matthias: This is quite a long/complicated name 😅 Do we need the `WithFullPath`?

Maybe `ModuleSymbols` for the static structs that are shared, then maybe `ProcessModuleSymbols` for the in-process loaded symbols with the load bias + path? 

Same for unwind data: We have static data which is shared across different processes (`UnwindDataV3`) and then specific data that depends on the process (just the load bias) which could be called `ProcessUnwindData` or `LoadedUnwindData`

src/executor/wall_time/perf/parse_perf_file.rs
```diff
@@ -167,29 +204,68 @@ fn process_mmap2_record(
...
+                debug!("Failed to compute load bias for {record_path_string}: {error}");
+            }
+        }
+    }
 
-    match UnwindData::new(
+    // NOTE: We still read the ELF file for every mapping even if we've already seen this path.
+    // The V3 data will be deduplicated below, but there's a possible optimization to skip
+    // the ELF read entirely when the path is already known, by caching the necessary data to
+    // compute load_bias.
```
not-matthias: side note: Since we're using memmap2 it might already be fast enough (except the copying of the DWARF data). but let's keep it in mind for the future

src/executor/wall_time/perf/perf_map.rs
```diff
@@ -126,17 +116,36 @@ impl ModuleSymbols {
             return Err(anyhow::anyhow!("No symbols found"));
         }
 
-        let load_bias = elf_helper::compute_load_bias(
+        Ok(Self {
+            load_bias: 0,
```
not-matthias: wait, shouldn't we remove this then?

src/bin/compare_walltime_output.rs
```diff
@@ -0,0 +1,900 @@
+/// Compares walltime profiling output between old and new runner formats.
```
not-matthias: I'm not sure if it make sense to keep this, since we won't need it anymore after this PR has been merged, right?

build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    // Force a rebuild of the test target to be able to run the full test suite locally just by
+    // setting GITHUB_ACTIONS=1 in the environment.
+    // This is because `test_with` is evaluated at build time
+    println!("cargo::rerun-if-env-changed=GITHUB_ACTIONS");
```
not-matthias, line 5: Do we really need this? I thought test_with has a runtime check

GuillaumeLagrange: yeah this is necessary since the test is actually at build time :nerd_face:

src/executor/wall_time/perf/parse_perf_file.rs
GuillaumeLagrange: Iv'e changed the whole logic, hopefully it's clearer

src/executor/wall_time/perf/perf_map.rs
GuillaumeLagrange: Yeah I removed it

src/bin/compare_walltime_output.rs
GuillaumeLagrange: Removed the script

src/executor/wall_time/perf/snapshots/codspeed_runner__executor__wall_time__perf__unwind_data__tests__cpp_unwind_data.snap
GuillaumeLagrange: Done

src/executor/wall_time/perf/mod.rs
```diff
@@ -4,45 +4,41 @@ use crate::cli::UnwindingMode;
...
 pub mod debug_info;
 pub mod elf_helper;
 pub mod fifo;
+pub mod module_symbols;
 pub mod perf_executable;
-pub mod perf_map;
 pub mod unwind_data;
 
-const PERF_METADATA_CURRENT_VERSION: u64 = 1;
+const PERF_METADATA_CURRENT_VERSION: u64 = 3;
```
not-matthias: Didn't we skip 2 here? 😄

src/executor/wall_time/perf/parse_perf_file.rs
```diff
@@ -1,19 +1,46 @@
...
 use linux_perf_data::linux_perf_event_reader::RecordType;
+use runner_shared::unwind_data::ProcessUnwindData;
 use runner_shared::unwind_data::UnwindData;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::path::Path;
+use std::path::PathBuf;
+
+#[derive(Default)]
+pub struct MountedModule {
```
not-matthias: Not sure if "mounted" is the right name for this. Wouldn't it be more like `MappedModule`?

src/executor/wall_time/perf/parse_perf_file.rs
```diff
@@ -1,19 +1,46 @@
...
+    /// Symbols extracted from the mapped ELF file
+    pub module_symbols: Option<ModuleSymbols>,
+    /// Unwind data extracted from the mapped ELF file
+    pub unwind_data: Option<UnwindData>,
+    /// Per-process mounting information
+    pub process_mounted_module: HashMap<pid_t, ProcessMountedModule>,
```
not-matthias: same here

src/executor/wall_time/perf/unwind_data.rs
```diff
@@ -145,13 +134,14 @@ mod tests {
 
         let (start_addr, end_addr, file_offset) =
             (0x0000000000402000_u64, 0x000000000050f000_u64, 0x2000);
-        assert_elf_load_bias!(start_addr, end_addr, file_offset, MODULE_PATH, 0x0);
-        insta::assert_debug_snapshot!(UnwindData::new(
+        let load_bias =
+            assert_and_get_load_bias(start_addr, end_addr, file_offset, MODULE_PATH, 0x0);
```
not-matthias: Don't have to change it, but we could instead also define a variable `load_bias` that is passed as the last parameter to `assert_load_bias`, which we then can also use below in the `unwind_data_from_elf`

Feels a bit cleaner than having a `assert_and_get` function

src/executor/wall_time/perf/parse_perf_file.rs
GuillaumeLagrange: Changed to `Loaded`, because I used `Mapped` to indicate the fact that we mapped the original binary file name to a deduplicated `path_key` in perf metadata

src/executor/wall_time/perf/save_artifacts.rs
```diff
@@ -0,0 +1,294 @@
...
+/// Register a path in the map if absent, assigning a new unique key.
+/// Returns the assigned key.
+fn get_or_insert_key(path_to_key: &mut HashMap<PathBuf, String>, path: &Path) -> String {
+    if let Some(key) = path_to_key.get(path) {
+        return key.clone();
+    }
+    let key = naming::indexed_semantic_key(path_to_key.len(), path);
+    path_to_key.insert(path.to_owned(), key.clone());
+    key
+}
```
art049, line 72: I think we loose a bit of info here. IIRC, we ended up saying that hashing the path was a better option no?
I know we iterated quite a lot but here we loose all info about similar prefixes (when in the same path) even if it doesn't happen often

GuillaumeLagrange: This hashmap holds the "indexed semantic key" for a given PathBuf (we ditched the hash in favor of the indexed_semantic_key here)

We then save the inverted hashmap (indexex semantic key -> pathbuf) in the perf.metadata, if we ever need to reconstruct the full path, so we are not losing information so far.

PR #241, opened 2026-02-12
feat: add tcmalloc support

not-matthias:


crates/memtrack/src/allocators/dynamic.rs
```diff
@@ -46,6 +46,25 @@ fn get_allocator_paths(lib: &AllocatorKind) -> &'static [&'static str] {
             // NixOS
             "/nix/store/*mimalloc*/lib/libmimalloc.so*",
         ],
+        AllocatorKind::Tcmalloc => &[
+            // gperftools tcmalloc variants
+            // Debian, Ubuntu: Standard Linux multiarch paths
+            "/lib/*-linux-gnu/libtcmalloc.so*",
```
GuillaumeLagrange: As discussed in my live podcast review

We could have some sort of regex template for each base distribution that points to the locations where libraries are to be looked for

Then for each alocator, we just least file names/subpaths that are relevant for the library, and perform the search for each distrib.

crates/memtrack/src/allocators/mod.rs
```diff
@@ -19,8 +19,15 @@ pub enum AllocatorKind {
     Jemalloc,
     /// mimalloc - Microsoft's allocator
     Mimalloc,
+    /// TCMalloc - Google's thread-caching malloc
+    ///
+    /// Two variants exist:
+    /// - **gperftools** (github.com/gperftools/gperftools): Original ~2005 release.
+    ///   Exports both standard symbols (malloc/free) AND tc_* prefixed symbols.
+    /// - **google/tcmalloc** (github.com/google/tcmalloc): Modern ~2020 rewrite.
+    ///   Exports ONLY standard symbols (malloc/free/etc.) - no tc_* prefix.
```
GuillaumeLagrange, line 28: This docstring mentions non `tc_` prefixed symbols, but in the symbols table bellow, they are not there.

Is this expected ? If no, let's fix, if yes, let's add a comment/docstring below to explain why we do not list the non prefixed symbols

crates/memtrack/testdata/alloc_cpp/cmake.toml
```diff
@@ -21,6 +21,53 @@ endif()
 if(NOT JEMALLOC_STATIC_LIB)
     message(STATUS "Static jemalloc not found, will fall back to dynamic linking")
 endif()
+
+# TCMalloc support via system gperftools package
+# gperftools provides libtcmalloc on most Linux distributions
+# (Ubuntu: libtcmalloc-minimal4, Fedora: gperftools-libs, NixOS: gperftools)
```
GuillaumeLagrange, line 27: we have a `flake.nix` at home 😢

crates/memtrack/src/allocators/mod.rs
not-matthias: non-tc prefixed symbols are basically libc-compatible allocation functions. the `.symbols()` is only required for statically linked allocators. and in those cases we'll simply detect it as `libc`. 

But you've got a point, we should try to also attach to libc functions for dynamic libraries (right now we only attach to libstdc++)

GuillaumeLagrange: My point was just to make it clear somewhere that the table below does not expect libc compatible symbols because we try to attach to libc symbols no matter what ( if I understood correctly)

not-matthias: Yep, we have two ways of detecting it: 
- Dynamically linked: Pattern matching from the path, we'll know exactly which allocator this is. Here we'll attach to liblc/libstdc++ no matter what. 
- Statically linked: Here we have the function name check; if we don't detect `tc_*` here, then we'll likely just detect it as `Libc` allocator.

I'll move the symbols() function to the static_linked to prevent confusion; and udate the docs string

crates/memtrack/testdata/alloc_cpp/cmake.toml
not-matthias: 😢

PR #261, opened 2026-03-09
feat: support perf walltime in Java

not-matthias:


src/executor/wall_time/perf/mod.rs
```diff
@@ -222,6 +222,16 @@ impl PerfRunner {
         let on_cmd = async |cmd: &FifoCommand| {
             #[allow(deprecated)]
             match cmd {
+                // Print /proc/{pid}/maps for the benchmark process. This helps with debugging missing symbols, as
+                // it allows finding the module in which an address is located.
+                FifoCommand::CurrentBenchmark { pid, .. } if is_codspeed_debug_enabled() => {
+                    let maps_path = format!("/proc/{pid}/maps");
+                    match std::fs::read_to_string(&maps_path) {
+                        Ok(maps) => debug!("/proc/{pid}/maps:\n{maps}"),
```
GuillaumeLagrange: I'm slightly confused by this, doesn't the perf file parsing print enough information ?
I'd rather we clean up/improve the logs in parse_perf_file's process_mmap2_record rather than doing this.

This is really limited because it only prints the self-reported pid of the benchmark rather than all relevant pids

src/executor/wall_time/perf/mod.rs
```diff
@@ -96,6 +96,12 @@ impl PerfRunner {
         // Infer the unwinding mode from the benchmark cmd
         let (cg_mode, stack_size) = if let Some(mode) = config.perf_unwinding_mode {
             (mode, None)
+        } else if config.command.contains("gradle")
+            || config.command.contains("java")
```
GuillaumeLagrange: we should be stricter here, because `javascript` inside a path to an executable will match this

GuillaumeLagrange: could be the occasion to create a dedicated helper to try and guess the language, then have a language to option translation, this way we can test the detection, as well as make the intent clearer on why for each language we want one option or the other.

I'm not happy with how we are doing the language detection yet we have to live with it until we improve it. However, the language to option is most likely here to stay so it deserves proper explanation

GuillaumeLagrange (changes requested):

src/executor/wall_time/perf/mod.rs
not-matthias: We now have a `detect_executable`. The trickiest part is handling all the different ways a command can be executed (e.g. concatenated cmds, path-referenced, etc.), so lmk if I missed something

src/executor/wall_time/perf/mod.rs
not-matthias: Perf parsing only shows modules that were loaded inside the memory. But Java (and other languages) sometimes have addresses located in RWX memory that's not backed by a module, which is why I added this. 

We can't do this in the perf parsing, since it doesn't contains this information. I can remove it and just keep it as an internal helper snippet though. We also can't print it at the end with all pids, since the process is not running anymore.

GuillaumeLagrange: isnt this a symptom of a deeper issue ? Like if the pid as mounted RWX that does not appear in the perf file, isnt what we are doing flawed ?

not-matthias: Not necessarily, as JIT-compiled languages will have this. Languages are then responsible to create a jitdump file for those RWX regions. We could print the ranges of those, but I think that'll be a bit too verbose since it's too fine-grained. 

I can remove the check and later just add it again in case we need it. It's not that complicated.

GuillaumeLagrange (approved):
lgtm

affinanti333 (commented):
I appreciate being part of the team

PR #280, opened 2026-03-29
feat: discover jitdump files

DaniPopes:
Jitdump discovery was hardcoded to `/tmp/jit-{pid}.dump`, missing dumps from LLVM ORC JIT (e.g. [revmc](https://github.com/paradigmxyz/revmc), Julia) which write to `~/.debug/jit/*/jit-{pid}.dump`.

Instead of searching hardcoded filesystem paths, extract jitdump file paths from MMAP2 records in the perf data, matching how `perf inject --jit` discovers them via `jit_detect()`.


not-matthias (changes requested):
Hey, it looks like the jitdump implementation was only done for the Walltime executor. As mentioned on Discord, we're currently in the process of switching to samply, so this change isn't really needed. 

Instead, the jitdump saving should be handled in src/executor/valgrind/executor.rs. Let me know if you have any questions.

DaniPopes:
woops

PR #286, opened 2026-04-12
feat(cli): add --ignore-failure flag (#242)

### LINKED ISSUE #242: Impossible to profile binary with exit status 1
My process exits with status 1, but I want to be able to profile its execution. `codspeed exec` fails in this case with:

```
Error: failed to execute the benchmark process: exit status: 1
```

It would be great to add a flag to allow non-zero exit codes, just like `hyperfine` does for example.

yaowubarbara:
Closes #242.

Adds `--ignore-failure` (short `-i`) to `codspeed exec`, inspired by hyperfine. When set, a non-zero exit status from the benchmarked command is logged as a warning instead of aborting the run.

**Reworked per review:** the flag now lives in the exec-harness crate (where the user command is directly executed), not in the executor layer. This correctly scopes the tolerance to the wrapped command exit code.

### Changes

**exec-harness** (`crates/exec-harness/`):
- Added `ignore_failure` field to `BenchmarkCommand` (flows via stdin JSON)
- Added `--ignore-failure` / `-i` CLI arg for direct invocation
- `walltime/benchmark_loop.rs`: warn + continue instead of bail on non-zero exit
- `analysis/mod.rs`: same for `perform()` and `perform_with_valgrind()`

**CLI** (`src/`):
- Added `--ignore-failure` to `ExecArgs` only (not shared, not on `codspeed run`)
- Added `ignore_failure: bool` to `BenchmarkTarget::Exec`
- Threaded through `multi_targets::build_exec_targets_pipe_command()` into the JSON payload

### Why `exec` only, not `run`?

`codspeed run` expects a command with built-in CodSpeed instrumentation (pytest-codspeed, cargo-codspeed, etc.). If that command fails, it's a real test failure. `codspeed exec` wraps arbitrary commands that may legitimately exit non-zero (interrupted servers, coverage harnesses, etc.).

### DIFF
--- a/crates/exec-harness/src/analysis/mod.rs
+++ b/crates/exec-harness/src/analysis/mod.rs
@@ -27,7 +27,14 @@ pub fn perform(commands: Vec<BenchmarkCommand>) -> Result<()> {
         let status = status.context("Failed to execute command")?;
 
         if !status.success() {
-            bail!("Command exited with non-zero status: {status}");
+            if benchmark_cmd.ignore_failure {
+                warn!(
+                    "Command exited with non-zero status: {status}; \
+                     continuing because --ignore-failure is set"
+                );
+            } else {
+                bail!("Command exited with non-zero status: {status}");
+            }
         }
 
         hooks.set_executed_benchmark(&name_and_uri.uri).unwrap();
@@ -68,7 +75,14 @@ pub fn perform_with_valgrind(commands: Vec<BenchmarkCommand>) -> Result<()> {
         bail_if_command_spawned_subprocesses_under_valgrind(child.id())?;
 
         if !status.success() {
-            bail!("Command exited with non-zero status: {status}");
+            if benchmark_cmd.ignore_failure {
+                warn!(
+                    "Command exited with non-zero status: {status}; \
+                     continuing because --ignore-failure is set"
+                );
+            } else {
+                bail!("Command exited with non-zero status: {status}");
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


GuillaumeLagrange:
@yaowubarbara thanks for the PR

I think the idea is fine, but we should not do it at the executor level (valgrind/ walltime)

As far as I see it, the _only_ use case of the command failing would be when running outside of a codspeed integration, i.e when going through [this](https://github.com/CodSpeedHQ/codspeed/blob/b3d5055ee3ba0d316ea021ff739c987dbe92862f/src/executor/config.rs#L20-L25)

That means that the allowing a non 0 exit code should be scoped to the exec-harness, so accept the cli flag only for `codspeed exec` and forwared it all the way to the exec harness.

Then [here](https://github.com/CodSpeedHQ/codspeed/blob/b3d5055ee3ba0d316ea021ff739c987dbe92862f/crates/exec-harness/src/walltime/benchmark_loop.rs#L29-L31) and [here](https://github.com/CodSpeedHQ/codspeed/blob/main/crates/exec-harness/src/analysis/mod.rs) we should not bail on a failure

yaowubarbara:
Thanks for the review @GuillaumeLagrange! That makes total sense — the executor should stay strict since it's running exec-harness (or the instrumented command), and the tolerance belongs inside exec-harness where the user's command is directly invoked.

Reworked:
- Removed ignore_failure from ExecutorConfig, OrchestratorConfig, ExecAndRunSharedArgs, and both executor impls
- Added it to BenchmarkTarget::Exec, flows through BenchmarkCommand JSON to exec-harness
- Scoped the CLI flag to codspeed exec only (via ExecArgs, not shared)
- The bail-to-warn change now lives in benchmark_loop.rs and analysis/mod.rs inside exec-harness

Force-pushed with the rework. Let me know if there's anything else to adjust!

PR #287, opened 2026-04-13
Support macos walltime without profiling in codspeed-runner

GuillaumeLagrange:


src/system/os.rs
```diff
@@ -0,0 +1,159 @@
...
+pub enum Os {
+    /// Linux distributions that support auto-install of CodSpeed tooling and are explicitly targeted by our executors
+    SupportedLinux {
+        distro: SupportedLinuxDistro,
+        version: String,
+    },
+    #[allow(clippy::enum_variant_names)]
+    MacOs { version: String },
+    /// Any other os we don't explicitly target. Executors may still work but the user is responsible for installing tooling themselves.
+    Other { name: String, version: String },
```
not-matthias, line 82: Isn't this kind of dangerous? For example, does Windows fall into this category? How about NixOS? 

Running on Windows should fail, while NixOS should work. We shouldn't merge both into one category.

src/system/os.rs
```diff
@@ -0,0 +1,159 @@
...
+/// Serialized via [`OsSerde`] as the flat `{os, osVersion}` pair the API expects, so a consumer
+/// can `#[serde(flatten)]` this field into a parent struct without altering the wire format.
+#[derive(Eq, PartialEq, Hash, Debug, Clone, Serialize, Deserialize)]
+#[serde(from = "OsSerde", into = "OsSerde")]
+pub enum Os {
+    /// Linux distributions that support auto-install of CodSpeed tooling and are explicitly targeted by our executors
+    SupportedLinux {
+        distro: SupportedLinuxDistro,
+        version: String,
+    },
```
not-matthias: How about renaming this to `Linux`, and including NixOS? We could have a `is_debian_based()` func on the LinuxDistro struct.

src/system/check.rs
```diff
@@ -1,56 +1,17 @@
...
+use super::{Os, SystemInfo};
 
-/// Checks if the provided system info is supported
-///
-/// Supported systems:
-/// - Ubuntu 20.04 x86_64
-/// - Ubuntu 22.04 x86_64 and aarch64
-/// - Debian 11 x86_64
-/// - Debian 12 x86_64
 pub fn check_system(system_info: &SystemInfo) -> Result<()> {
```
not-matthias: We could remove this file entirely and create `Os::is_supported`, or maybe have a SupportLevel (supported, best-effort, unsupported)

src/executor/valgrind/executor.rs
```diff
@@ -19,8 +19,16 @@ impl Executor for ValgrindExecutor {
...
 
-    fn tool_status(&self) -> ToolStatus {
-        get_valgrind_status()
+    fn tool_status(&self) -> Option<ToolStatus> {
+        Some(get_valgrind_status())
+    }
+
+    fn support_level(&self, os: &Os) -> ExecutorSupport {
+        match os {
+            Os::SupportedLinux { .. } => ExecutorSupport::FullySupported,
```
not-matthias: Shouldn't we check the version here since we have this download logic: 

```rust
    let (deb_ubuntu_version, architecture) =
        match (distro, version.as_str(), system_info.arch.as_str()) {
            (SupportedLinuxDistro::Ubuntu, "22.04", "x86_64")
            | (SupportedLinuxDistro::Debian, "12", "x86_64") => ("22.04", "amd64"),
            (SupportedLinuxDistro::Ubuntu, "24.04", "x86_64") => ("24.04", "amd64"),
            (SupportedLinuxDistro::Ubuntu, "22.04", "aarch64")
            | (SupportedLinuxDistro::Debian, "12", "aarch64") => ("22.04", "arm64"),
            (SupportedLinuxDistro::Ubuntu, "24.04", "aarch64") => ("24.04", "arm64"),
            _ => bail!("Unsupported system"),
        };

```

src/system/os.rs
```diff
@@ -0,0 +1,159 @@
...
+    MacOs { version: String },
+    /// Any other os we don't explicitly target. Executors may still work but the user is responsible for installing tooling themselves.
+    Other { name: String, version: String },
+}
+
+impl Os {
+    /// Build an [`Os`] from the raw `(os, version)` strings reported by `sysinfo`.
+    ///
+    /// Unknown Linux-ish distro ids fall into [`Os::OtherLinux`]; only OSes we know we cannot run
+    /// on (today: Windows) map to [`Os::Unsupported`].
```
not-matthias: Os::Unsupported and Os::OtherLinux dont exist

src/system/check.rs
not-matthias: Actually, it might be better to iterate over all executors and see if any are supported.

not-matthias (changes requested):
It's important to consider everything unsupported by default. e.g. if someone runs on a system we haven't considered (e.g. Windows) we shouldn't even try to continue running with best-effort.

Logic could be implemented like this:
- Tool support: Get `Os` struct, check version + distro 
- Executor: Check each OS, e.g. on Linux: is_debian FullySupported; else RequireManualInstallation; Mac/Windows: Unsupported
- General (check_system): We could refactor this to just iterate over all executors and see if any are supported.

GuillaumeLagrange:
@not-matthias I've done the following

- `Os` -> `SupportedOs`, we now can represent ONLY the Os that we at least somewhat support, and bail for the others
- The `SupportedOs` has two variants, Linux and MacOs, with Linux having a linux distrib enum for well known distirbs that we support better
- Updated the support level check for valgrind executor

src/executor/valgrind/executor.rs
```diff
@@ -19,8 +19,21 @@ impl Executor for ValgrindExecutor {
...
-    fn tool_status(&self) -> ToolStatus {
-        get_valgrind_status()
+    fn tool_status(&self) -> Option<ToolStatus> {
+        Some(get_valgrind_status())
+    }
+
+    fn support_level(&self, system_info: &SystemInfo) -> ExecutorSupport {
+        match &system_info.os {
+            SupportedOs::Linux(_) => {
+                if get_codspeed_valgrind_filename(system_info).is_ok() {
```
not-matthias: nit: was a bit confused that we're trying to get the valgrind filename here. maybe extract the support status into a separate function

src/system/check.rs
```diff
@@ -1,56 +1,8 @@
...
-    });
-
-/// Checks if the provided system info is supported
-///
-/// Supported systems:
-/// - Ubuntu 20.04 x86_64
-/// - Ubuntu 22.04 x86_64 and aarch64
-/// - Debian 11 x86_64
-/// - Debian 12 x86_64
 pub fn check_system(system_info: &SystemInfo) -> Result<()> {
```
not-matthias: do we still need this func/file? we could move the log and remove it entirely

not-matthias (approved):
LGTM, just two minor style comments

src/system/check.rs
GuillaumeLagrange: no I thought I removed it, thanks!

PR #300, opened 2026-04-17
feat: display a warning when profiling generation failed

GuillaumeLagrange:


src/upload/poll_results.rs
```diff
@@ -204,7 +217,20 @@ async fn display_single_run_results(
     Ok(())
 }
 
+fn warn_callgraph_failures(names: &[&str]) {
+    if names.is_empty() {
+        return;
+    }
+    warn!("We were unable to generate profiling data for the following benchmarks:");
```
not-matthias, line 224: Should we also print mitigations here? e.g. "try running the benchmark again" or give specific feedback based on the error. Otherwise agents might try random things that will not work

not-matthias (approved):

src/upload/poll_results.rs
GuillaumeLagrange: I'm not really happy with mitigations proposed in the frontend today, so I think it's not worth it to do here

PR #303, opened 2026-04-20
fix: missing libc symbols

not-matthias:


testdata/perf_map/cpp_my_benchmark.debug
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
GuillaumeLagrange, line 1: For this test, couldnt we just juste the actual libc binaries from ubuntu packages rather than artificially creating them?

src/executor/wall_time/perf/elf_helper.rs
```diff
@@ -188,3 +190,123 @@ pub fn relative_address_base(object_file: &object::File) -> u64 {
...
     base_svma.wrapping_add(load_bias)
 }
+
+const DEFAULT_DEBUG_DIR: &str = "/usr/lib/debug";
+
+/// Search for a separate debug info file.
+///
+/// Tries two mechanisms in order:
+/// 1. **Build-ID path**: `<debug_dir>/.build-id/<XX>/<YYYYYY...>.debug`
+/// 2. **`.gnu_debuglink`** with GDB search order and CRC32 validation
```
GuillaumeLagrange, line 200: Why this order? Instinctively I would look for whatever is directly embedded in the binary, but we could add a small sentence to justify the choice here. I'm guessing this is what matches first in ubuntu

src/executor/wall_time/perf/elf_helper.rs
```diff
@@ -188,3 +190,123 @@ pub fn relative_address_base(object_file: &object::File) -> u64 {
...
+        return Some(path);
+    }
+    find_debug_file_by_debuglink(object, binary_path, debug_dir)
+}
+
+fn find_debug_file_by_build_id(object: &object::File, debug_dir: &Path) -> Option<PathBuf> {
+    let build_id = object.build_id().ok()??;
+    if build_id.len() < 2 {
+        return None;
+    }
```
GuillaumeLagrange: What's this?

src/executor/wall_time/perf/elf_helper.rs
GuillaumeLagrange: To clarify: we should do what works best on ubuntu, and libc binaries from ubuntu repositories, just explain the order here

src/executor/wall_time/perf/module_symbols.rs
```diff
@@ -82,6 +81,44 @@ impl ModuleSymbols {
...
+            if let Some(debug_symbols) = debug_symbols {
+                let existing: HashSet<(u64, String)> =
+                    symbols.iter().map(|s| (s.addr, s.name.clone())).collect();
+                symbols.extend(
+                    debug_symbols
+                        .into_iter()
+                        .filter(|s| !existing.contains(&(s.addr, s.name.clone()))),
+                );
+            }
+        }
```
GuillaumeLagrange, line 120: If we have found symbols in the original elf, is there really a point to look for additional symbols in the debug file?
For debug infos, we do not do this if we found anything

GuillaumeLagrange (approved):
olgtm

src/executor/wall_time/perf/elf_helper.rs
not-matthias: The main reason is that build-id lookups are cheaper + more modern. I'll add a comment 👍

src/executor/wall_time/perf/elf_helper.rs
not-matthias: Just a sanity check that the debugid is valid since we take the first 2 chars and use it for the folder name, then the  remaining are the filename. I'll add a comment to explain this better

src/executor/wall_time/perf/module_symbols.rs
not-matthias: Yes, it's really important that we do this beacuse there are still symbols inside a stripped binary (e.g. any exported function).

PR #304, opened 2026-04-20
fix: update setarch command arguments to use long options

xtqqczze:
Replace the short `-R` option with the more explicit `--addr-no-randomize` option to improve clarity.

This change is intended to make `@codspeedbot` output easier to interpret:

https://github.com/uutils/coreutils/pull/11908#issuecomment-4281915697


art049 (approved):
Thanks

PR #310, opened 2026-04-22
feat: warn when libc debug info is not found

not-matthias:
When a stripped libc is loaded without a matching debug file (no .debug_info section and no match via build-id or .gnu_debuglink), emit a warning pointing users at libc6-dbg / glibc-debuginfo so they know why libc frames show as unknown symbols.

### DIFF
--- a/src/executor/wall_time/perf/debug_info.rs
+++ b/src/executor/wall_time/perf/debug_info.rs
@@ -60,12 +60,13 @@ impl ModuleDebugInfoExt for ModuleDebugInfo {
         let ctx = if object.section_by_name(".debug_info").is_some() {
             Self::create_dwarf_context(&object).context("Failed to create DWARF context")?
         } else {
-            let debug_path = find_debug_file(&object, path.as_ref()).with_context(|| {
-                format!(
+            let Some(debug_path) = find_debug_file(&object, path.as_ref()) else {
+                warn_missing_libc_debug_info(path.as_ref());
+                anyhow::bail!(
                     "No DWARF in {:?} and no separate debug file found",
                     path.as_ref()
-                )
-            })?;
+                );
+            };
             trace!(
                 "Using separate debug file {debug_path:?} for {:?}",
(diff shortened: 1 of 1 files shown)


src/executor/wall_time/perf/debug_info.rs
```diff
@@ -129,6 +130,23 @@ impl ModuleDebugInfoExt for ModuleDebugInfo {
     }
 }
 
+fn warn_missing_external_debug_info(path: &Path) -> Option<()> {
+    let file_name = path.file_name()?.to_str()?;
+    if !file_name.ends_with(".so") {
```
GuillaumeLagrange: can't there be some prefixes in .so files? like .so.6 or sthing? We'd never trigger this

src/executor/wall_time/perf/debug_info.rs
```diff
@@ -129,6 +130,23 @@ impl ModuleDebugInfoExt for ModuleDebugInfo {
...
+        return None;
+    }
+
+    if file_name.starts_with("libc.so") || file_name.starts_with("libc-") {
```
GuillaumeLagrange: The function name should convey that we are ONLY doing it for glibc, so it should be warn_missing_libc_debug_info

src/executor/wall_time/perf/debug_info.rs
```diff
@@ -129,6 +130,23 @@ impl ModuleDebugInfoExt for ModuleDebugInfo {
...
+        warn!(
+            "Debug info for {} not found. Install libc6-dbg (Debian/Ubuntu) or \
+             glibc-debuginfo (Fedora/RHEL) to avoid fix missing symbols in the flamegraph",
```
GuillaumeLagrange: typo here as mentioned by copilot

GuillaumeLagrange (changes requested):
small comments

GuillaumeLagrange (approved):
meant to approve

PR #313, opened 2026-04-24
feat(auth): add `--with-token` flag to read token from stdin

fargito:
Add ability to run `codspeed auth login --with-token`, reading the token from stdin instead of performing the normal login flow.

Example: `echo "co1_xxxxxxx" | codspeed auth login --with-token`

This matches the behavior of the `gh` CLI: https://cli.github.com/manual/gh_auth_login

### DIFF
--- a/src/cli/auth.rs
+++ b/src/cli/auth.rs
@@ -1,3 +1,4 @@
+use std::io::Read;
 use std::time::Duration;
 
 use crate::api_client::{CodSpeedAPIClient, GetRepositoryVars};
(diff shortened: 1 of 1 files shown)


src/cli/auth.rs
```diff
@@ -33,54 +38,70 @@ pub async fn run(
...
-    if open::that(&login_session_payload.callback_url).is_ok() {
-        info!("Your browser has been opened to complete the login process");
+    let token = if with_token {
+        let mut buf = String::new();
+        std::io::stdin().read_to_string(&mut buf)?;
+        let token = buf.trim().to_owned();
+        if token.is_empty() {
+            bail!("No token provided on stdin");
+        }
+        token
```
GuillaumeLagrange, line 63: Shouldnt we check for token validity before writing it to the config file?

GuillaumeLagrange (approved):
olgtm

src/cli/auth.rs
fargito: I'm not sure. We can use `codspeed auth status` afterwards, it's simpler for now.

fargito: Otherwise we'd have to instantiate a client without reading the config from file, which is too cumbersome for the added value IMO

PR #314, opened 2026-04-24
feat: only wrap walltime command with sudo on linux

GuillaumeLagrange:
We do not yet need sudo, since we don't use launchtl nor profilers yet.

image


not-matthias (approved):
LGTM!

PR #327, opened 2026-04-30
feat(system): detect ubuntu/debian derivative distros via id_like fallback

### LINKED ISSUE #326: Valgrind setup fails on Ubuntu/Debian derivative distributions (e.g. Pop!_OS)
## Context

https://github.com/CodSpeedHQ/codspeed/blob/a7eb6e2ca7de8308f0d7203aea2abef0399d2f94/src/system/os.rs#L24-L35

Running `cargo test` on Ubuntu-derivative distros (e.g. Pop!_OS for me) causes 14 valgrind executor tests to fail with: `Unsupported system`.

```bash
test executor::tests::valgrind::test_valgrind_executor_with_env::case_2 ... FAILED
test executor::tests::valgrind::test_valgrind_executor_with_env::case_3 ... FAILED
test executor::tests::valgrind::test_valgrind_executor_with_env::case_4 ... FAILED
AGENTS.md   Cargo.toml	  CONTRIBUTING.md      LICENSE-APACHE	    schemas  target
assets	    CHANGELOG.md  crates	       LICENSE-MIT	    scripts  testdata
build.rs    CLAUDE.md	  cspell.json	       README.md	    skills   tests
Cargo.lock  cliff.toml	  dist-workspace.toml  rust-toolchain.toml  src
test executor::tests::valgrind::test_valgrind_executor_with_env::case_5 ... FAILED
test executor::shared::fifo::tests::recv_cmd_is_not_cancel_safe ... ok
test executor::tests::valgrind::test_valgrind_executor_with_env::case_6 ... FAILED
test executor::tests::valgrind::test_valgrind_executor_with_env::case_7 ... FAILED
test executor::wall_time::perf::elf_helper::tests::test_find_debug_file_by_build_id ... ok
test executor::tests::valgrind::test_valgrind_executor_with_env::case_8 ... FAILED\
```

Since Pop!_OS is a derivative of `ubuntu|debian`, I though we could adapt the process to accept other distros that are close to `Ubuntu` and `Debian`

## Root cause

`sysinfo::System::distribution_id()` returns the distro's own ID (e.g. `"pop"` for Pop!_OS), not its parent. `LinuxDistribution::from_id()` only exact-matches `"ubuntu"` and `"debian"`, so derivatives fall through to `Other`. Which `get_codspeed_valgrind_filename()` doesn't handle.

## Proposal
Use `System::distribution_id_like()` (available since `sysinfo@0.35 I think) as a fallback: if the primary ID is unrecognized, walk the parent chain (`["ubuntu", "debian"]` for Pop!_OS) until a known distro is found.

https://docs.rs/sysinfo/0.38.4/sysinfo/struct.System.html#method.distribution_id_like

This requires bumping `sysinfo` from `0.33.1` to `0.38.4`, which also involves adapting one call site where `physical_core_count()` became an associated function.

## Affected systems

Any Linux distro where `ID != ubuntu|debian` but `ID_LIKE` contains one of them.

s3bc40:
Closes #326

## Summary

- Upgrade `sysinfo` to `0.38.4` to access `System::distribution_id_like()`
- Add `LinuxDistribution::from_id_like()` which falls back to the `ID_LIKE` parent chain when the primary distro ID is unrecognized
- Adapt `System::physical_core_count()` call site in `system/info.rs` (became an associated function in `0.38`)

## Test plan

- `pop_os_resolves_to_ubuntu_via_id_like` — primary fix case
- `ubuntu_resolves_directly_without_id_like` — existing behavior unchanged
- `centos_with_unrecognized_parents_is_other` — unrecognized lineage stays `Other`
- All existing tests pass

---

ofc if it does not comply with your needs, you can close the PR 👍

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -734,7 +734,7 @@ dependencies = [
  "shellexpand",
  "simplelog",
  "strum",
- "sysinfo 0.33.1",
+ "sysinfo 0.38.4",
  "tabled",
  "temp-env",
  "tempfile",
@@ -4159,21 +4159,6 @@ dependencies = [
 
 [[package]]
 name = "sysinfo"
-version = "0.33.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4fc858248ea01b66f19d8e8a6d55f41deaf91e9d495246fd01368d99935c6c01"
-dependencies = [
- "core-foundation-sys",
- "libc",
- "memchr",
- "ntapi",
- "rayon",
- "serde",
- "windows 0.57.0",
-]
-
-[[package]]
-name = "sysinfo"
 version = "0.35.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3c3ffa3e4ff2b324a57f7aeb3c349656c7b127c3c189520251a648102a92496e"
@@ -4184,6 +4169,21 @@ dependencies = [
  "objc2-core-foundation",
  "objc2-io-kit",
  "windows 0.61.3",
+]
+
+[[package]]
+name = "sysinfo"
+version = "0.38.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92ab6a2f8bfe508deb3c6406578252e491d299cbbf3bc0529ecc3313aee4a52f"
+dependencies = [
+ "libc",
+ "memchr",
+ "ntapi",
+ "objc2-core-foundation",
+ "objc2-io-kit",
+ "serde",
+ "windows 0.62.2",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


GuillaumeLagrange:
Hello!

For now we have not tested the auto installation of tools on distributions other than ubuntu and debian, and I fear we might run into half-working behaviors if we just merge this.

I did not know about the ID_LIKE spec, but I do not think it is worth it to manage this for now, as it would require checking that the `.deb`s that we generate in https://github.com/CodSpeedHQ/valgrind-codspeed are functional as soon as the ID_LIKE is `debian`, which I doubt.

For now, the recommandation is to manually install the required tools by following instructions here: https://codspeed.io/docs/cli#manual-tool-installation

s3bc40:
Understandable, it's not like a priority. Even if these distros are from the same family, we never know. There might be a slight difference that might bring issues.

I'll follow along the manual install doc and redo the tests as you suggested. Thanks for the feedback, i can close the PR and solve the issue if you want :+1:

GuillaumeLagrange:
I'll close this for now because since we have the manual installation instructions we'd like to avoid the maintenance burden of making sure distro edge cases work etc, but we'll keep it in mind as the cli evolves.

Thanks for the PR/input anyway, always appreciated!

s3bc40:
Fine by me! Thanks for the review and your explanation.

PR #334, opened 2026-05-04
feat: add a way to disable perf compression through env var

GuillaumeLagrange:
This is aimed to be dev only, because some profilers do not support compressed perf data, namely hotspot.

### DIFF
--- a/src/executor/wall_time/perf/perf_executable.rs
+++ b/src/executor/wall_time/perf/perf_executable.rs
@@ -134,6 +134,10 @@ pub fn get_compression_flags<S: AsRef<Path>>(perf_executable: S) -> Result<Optio
 
     if has_zstd {
         debug!("perf supports zstd compression");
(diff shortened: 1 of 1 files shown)


not-matthias (approved):

src/executor/wall_time/perf/perf_executable.rs
```diff
@@ -134,6 +134,10 @@ pub fn get_compression_flags<S: AsRef<Path>>(perf_executable: S) -> Result<Optio
 
     if has_zstd {
         debug!("perf supports zstd compression");
+        if std::env::var("CODSPEED_DISABLE_PERF_COMPRESSION").is_ok() {
+            info!("CODSPEED_DISABLE_PERF_COMPRESSION is set, disabling perf compression");
```
not-matthias: nit: i think our other env vars have the `CODSPEED_PERF` format, so we could name this `CODSPEED_PERF_DISABLE_COMPRESSION`

PR #337, opened 2026-05-05
chore: increase stack sampling for python

GuillaumeLagrange:
mitigates unwinding issues that now impact the flamegraph's topology much more than before since we do not merge simply by symbol name


not-matthias (approved):

not-matthias:
make sure to test the changes on pytest-codspeed (we previously had issues with large perf.data files)

PR #338, opened 2026-05-05
feat: add samply profiler support

not-matthias:


GuillaumeLagrange (changes requested):
Switching to github mid review, have to publish 😬

src/executor/wall_time/profiler/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// on `self`.
+    async fn wrap(
+        &mut self,
+        cmd: CommandBuilder,
+        config: &ExecutorConfig,
+        profile_folder: &Path,
+    ) -> anyhow::Result<CommandBuilder>;
+
+    /// The benchmarked process signaled the start of a measured region.
+    async fn on_start_benchmark(&mut self) -> anyhow::Result<()> {
```
GuillaumeLagrange: Could we make use of the fact that we are changing stuff here to clear up the name of the markers?<br>We've discussed this multiple time, and it's quite a huge clusterfuck

The current names with markers is this

```
    /// It expects the data to be structured like this:
    /// - SampleStart
    ///    - BenchmarkStart
    ///       - N samples
    ///    - BenchmarkEnd
    /// - SampleEnd
```

But we also have the fifo commands `StartBenchmark` and `StopBenchmark` which actually [push `SampleStart/End` markers](<https://github.com/CodSpeedHQ/codspeed/blob/7d22fc9d2339ed738784d2c5cea441c5e38416d4/src/executor/shared/fifo.rs#L210-L218>)

What I propose for this is just a rename that would make everything a bit clearer

The external marker, currently named `SampleStart` should be named `ProfilerSamplingStart`, or something that conveys the idea that we are starting and stopping the profiler's sampling functionality, and we should rename the `FifoCommand::StartBenchmark` to reflect that. Another naming possibility could be `BenchmarkRegion` if we want to be a bit farther from the implementation detail of starting/stopping the profiler

The internal marker would be clearer if it was something like `RoundStart` `RoundEnd`, as in practice it's what is being done.

WE SHOULD BE ABSOLUTELY EXTREMELY CAREFUL TO MAKE SURE IT DOES NOT IMPACT THE SERIALIZED IDs, it's only meant to be a naming clarification.

WDYT?

src/executor/wall_time/perf/mod.rs
```diff
@@ -240,7 +240,7 @@ impl PerfRunner {
                 }
                 FifoCommand::GetIntegrationMode => {
                     return Ok(Some(FifoCommand::IntegrationModeResponse(
-                        IntegrationMode::Perf,
+                        IntegrationMode::Walltime,
```
GuillaumeLagrange: We should make absolutely sure that the ser/de test inside instrument-hooks still works, and I believe explicitly updated.

Not sure which is the best order, but we need to be careful this does not get left behind as it had been before

src/executor/wall_time/profiler/perf/mod.rs
```diff
@@ -28,18 +28,13 @@ use std::path::Path;
 use std::path::PathBuf;
```
GuillaumeLagrange, line 31: Seen together, this commit is pretty much useless since smaply handles the symbolification

src/executor/wall_time/executor.rs
```diff
@@ -72,14 +77,17 @@ impl Drop for HookScriptsGuard {
 }
 
 pub struct WallTimeExecutor {
-    perf: Option<PerfRunner>,
+    profiler: Option<Box<dyn Profiler>>,
```
GuillaumeLagrange, line 84: Since it wont really change at runtime and is computed from static cfg asserts anyway, wouldnt it be better to have a type generic profiler rather than a `Box<dyn>` ?

src/executor/wall_time/executor.rs
```diff
@@ -187,13 +193,6 @@ impl Executor for WallTimeExecutor {
     async fn teardown(&self, execution_context: &ExecutionContext) -> Result<()> {
         debug!("Copying files to the profile folder");
 
-        if let Some(perf) = &self.perf
-            && execution_context.config.enable_perf
-        {
-            perf.save_files_to(&execution_context.profile_folder)
```
GuillaumeLagrange, line 193: Is there a reason the `finalize` part moved from teardown step to the `execute` step?

I'm guessing it has no side effect, but still conceptually it makes more sense to me to have the perf parsing etc as part of the executor teardown rather than inside the run step. WDYT?

src/executor/wall_time/executor.rs
```diff
@@ -136,33 +142,33 @@ impl Executor for WallTimeExecutor {
...
-                && execution_context.config.enable_perf
-            {
-                perf.run(
+        let _guard = HookScriptsGuard::setup();
+
+        let (_env_file, _script_file, cmd_builder) =
+            WallTimeExecutor::walltime_bench_cmd(&execution_context.config, execution_context)?;
+
+        let status = match self.profiler.as_mut() {
+            Some(profiler) if execution_context.config.enable_perf => {
```
GuillaumeLagrange: Ignore if it is addressed in later commits, but we should rename everything that is perf related to profiler in the CLI interface, but still accept the perf options with a dperecation message

GuillaumeLagrange: 👍

src/cli/shared.rs
```diff
@@ -152,16 +152,34 @@ pub enum UnwindingMode {
 
 #[derive(Args, Debug, Clone)]
 pub struct PerfRunArgs {
```
GuillaumeLagrange: this should be renamed and the unwinding mode argument should be scoped to perf, could be done using a generic to avoid having to handle this at runtime

src/executor/shared/fifo.rs
```diff
@@ -177,9 +176,37 @@ impl RunnerFifo {
...
+                time * nanos_per_tick.numer as u64 / nanos_per_tick.denom as u64
+            }
+
+            #[cfg(not(target_os = "macos"))]
+            {
+                use nix::{sys::time::TimeValLike, time::clock_gettime};
+
+                let clock = nix::time::ClockId::CLOCK_MONOTONIC;
+                clock_gettime(clock).unwrap().num_nanoseconds() as u64
+            }
```
GuillaumeLagrange: Maybe stupid question: wouldnt it be better to use the instrument_hooks wrapper crate that we have in the repo already?

GuillaumeLagrange: Would reduce teh chacnes of failure in the future, and get rid of the mach2 dependency

src/executor/wall_time/profiler/samply/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+use runner_shared::artifacts::ArtifactExt;
+use runner_shared::artifacts::ExecutionTimestamps;
+use runner_shared::metadata::WalltimeMetadata;
+use std::path::Path;
+use std::path::PathBuf;
+
+use super::WALLTIME_METADATA_CURRENT_VERSION;
+
+const SAMPLY_OUTPUT_FILE_NAME: &str = "samply-profile.json.gz";
+const SAMPLY_RATE_HZ: &str = "997";
```
GuillaumeLagrange: Let's have this in common as part of the `profiler` module, so that they dont drift without us being aware in the future

src/executor/wall_time/profiler/samply/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl SamplyProfiler {
+    pub fn new() -> Self {
+        Self { output_path: None }
+    }
+}
+
+#[async_trait(?Send)]
+impl Profiler for SamplyProfiler {
+    async fn wrap(
```
GuillaumeLagrange: can we rename this to `wrap_command` to make it slightly clearer?

src/executor/wall_time/profiler/samply/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    async fn finalize(
+        &mut self,
+        fifo_data: FifoBenchmarkData,
+        timestamps: ExecutionTimestamps,
+        profile_folder: &Path,
+    ) -> anyhow::Result<()> {
+        let Some(integration) = fifo_data.integration.clone() else {
+            warn!(
+                "Walltime profiling is enabled, but failed to detect benchmarks. If you wish to disable this warning, set CODSPEED_PROFILER_ENABLED=false"
```
GuillaumeLagrange: Nit: could be commonized with perf error message

GuillaumeLagrange: potentially through a trait function, called inside `finalize`, since the detection condition seem extremely similar

GuillaumeLagrange: but it may be overkill, at least just commonize the message

src/cli/mod.rs
```diff
@@ -117,6 +118,9 @@ enum Commands {
     Show,
     /// Update the CodSpeed CLI to the latest version
     Update,
+    /// Run the bundled samply profiler. Args are forwarded to samply.
+    #[command(disable_help_flag = true, disable_help_subcommand = true)]
+    Samply(samply::SamplyArgs),
 }
```
GuillaumeLagrange, line 133: Can we segregate these subcommands into a dedicated enum so we now at glance which commands are real entrypoints by users and which are actually just the cli re-invoking itself?

src/executor/wall_time/profiler/samply/mod.rs
```diff
@@ -50,8 +50,13 @@ impl Profiler for SamplyProfiler {
     ) -> anyhow::Result<CommandBuilder> {
         let output_path = profile_folder.join(SAMPLY_OUTPUT_FILE_NAME);
 
-        let mut samply_builder = CommandBuilder::new("samply");
+        // samply is bundled into this binary as the `samply` subcommand;
+        // re-exec ourselves so we don't depend on a system install.
+        let current_exe = std::env::current_exe()
```
GuillaumeLagrange: If we define the subcommands that are invoked by the CLI, we could have this as a helper that would just take the enum value and build the command with args etc

src/cli/samply.rs
```diff
@@ -16,16 +16,22 @@ pub fn run(args: SamplyArgs) -> Result<()> {
...
-        #[cfg(any(
-            target_os = "android",
-            target_os = "macos",
-            target_os = "linux",
-            target_os = "windows"
-        ))]
-        cli::Action::Record(a) => ::samply::do_record_action(a),
-        _ => unimplemented!("Only `samply record` is supported"),
-    }
+    // samply spins up its own tokio runtime internally, so it must run on a
```
GuillaumeLagrange, line 19: fixup with previous commit

GuillaumeLagrange (changes requested):
first pass done, let me know if you want to sync on some points

src/executor/wall_time/profiler/samply/mod.rs
not-matthias: I'm not quite sure I get what you mean, because we need to pass it as a string to be able to run `codspeed samply ...`. Could you give me an exaxmple?

src/executor/wall_time/profiler/mod.rs
not-matthias: Wdyt about this: 

```
pub enum MarkerType {
    SampleStart(u64),
    SampleEnd(u64),
    RoundStart(u64),
    RoundEnd(u64),
}
```

and then rename the commands to `StartProfiler/StopProfiler/PingProfiler`

src/executor/wall_time/perf/mod.rs
not-matthias: https://github.com/CodSpeedHQ/instrument-hooks/pull/27

src/executor/wall_time/executor.rs
not-matthias: We could, but then we'd have to represent the case where we dont' have a profiler (e.g. on Windows). We could have a NoopProfiler that does nothing, but I don't feel like this properly fits what we want to represent. 

And the overhead of Box<dyn> shouldn't be an issue because we're not really calling this in a hot path where the performance matters. So I'd leave it as-is, lmk if you disagree

GuillaumeLagrange: Ok let's keep things simple

src/executor/wall_time/profiler/samply/mod.rs
GuillaumeLagrange: Discussed it IRL, let's have a base `.get_command_builder` defined on `InternalCommands` so that each enum variant is allowed to decide how it invokes the underlying tool

.github/workflows/ci.yml
```diff
@@ -80,6 +80,8 @@ jobs:
           cargo install --path crates/exec-harness --locked
 
       - name: Run tests
+        env:
+          CODSPEED_PROFILER_ENABLED: "false"
```
GuillaumeLagrange, line 85: add a small comment explaining that we do not support profiling system commands with samply yet

src/executor/helpers/env.rs
```diff
@@ -24,7 +24,8 @@ pub fn get_base_injected_env(
         ("PYTHONHASHSEED".into(), "0".into()),
         (
             "PYTHON_PERF_JIT_SUPPORT".into(),
-            if mode == RunnerMode::Walltime {
+            // IMPORTANT: We must not enable this, otherwise we'll have many unresolved addresses on the stack on MacOS
```
GuillaumeLagrange:
```
// IMPORTANT: Keep this disabled on MacOS
```

+ create an issue

src/executor/shared/fifo.rs
```diff
@@ -161,11 +175,9 @@ impl RunnerFifo {
...
         let mut integration = None;
 
-        let current_time = || {
-            clock_gettime(nix::time::ClockId::CLOCK_MONOTONIC)
-                .unwrap()
-                .num_nanoseconds() as u64
-        };
+        // Must match the clock used by the benchmarked process so timestamps
+        // from both sides are comparable.
+        let current_time = instrument_hooks_bindings::InstrumentHooks::current_timestamp;
```
GuillaumeLagrange: nitpick: `get_current_time` because it's slightly less clear while reading that it's a function now

GuillaumeLagrange (approved):
olgtm

PR #341, opened 2026-05-07
feat: validate tokens and repository access up front

GuillaumeLagrange:
Catch invalid or mis-scoped tokens before any benchmarks run, instead of
surfacing a 401 from the upload endpoint after a full suite has executed.

The local provider now validates the token and the repository's
`CREATE_LOCAL_RUN` access in a single GraphQL round-trip when resolving
from a `-r` override or a detected git remote. The project-repository
fallback is unchanged. `auth status` uses the same combined query so it
doesn't double-roundtrip when a remote is detected, and `auth login`
validates the token via `session()` before persisting it.

The bulk of the diff is a prerequisite refactor: the api_client is now
the single source of truth for the auth token. Previously the token
lived in two places — on the api_client for GraphQL `Authorization`
headers, and on `OrchestratorConfig`/`ExecutorConfig` for upload
`Authorization` headers — and the two could drift, notably in CI with
OIDC where `set_oidc_token` mutated config but the api_client kept its
original token. `CodSpeedAPIClient` now owns the token through `token()`
and `set_token()`, the trait method becomes
`refresh_token(&mut CodSpeedAPIClient)`, and `&mut api_client` is
plumbed through `cli/run/exec → Orchestrator → upload_all`. The token
field is deleted from both configs.

Reviewers: the two commits are independently reviewable. The first
(`ref: make api_client the single source of truth…`) is a pure
refactor with no behavior change. The second (`feat: validate
tokens…`) is the user-facing change.

Refs COD-2628

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ serde = { workspace = true }
 serde_json = { workspace = true, features = ["preserve_order"] }
 url = "2.4.1"
 sha256 = "1.4.0"
-tokio = { version = "1", features = ["macros", "rt"] }
+tokio = { version = "1", features = ["macros", "rt", "signal"] }
 tokio-tar = { package = "astral-tokio-tar", version = "0.6.0" }
 tokio-util = "0.7.16"
 md5 = "0.7.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ simplelog = { version = "0.12.1", default-features = false, features = [
 tempfile = { workspace = true }
 git2 = "0.20.4"
 nestify = "0.3.3"
-gql_client = { git = "https://github.com/CodSpeedHQ/gql-client-rs" }
+gql_client = { git = "https://github.com/CodSpeedHQ/gql-client-rs", rev = "617a889e1d9089b5aa36679f1a5aa13ad01993ae" }
 serde_yaml = "0.9.34"
 sysinfo = { version = "0.33.1", features = ["serde"] }
 indicatif = "0.17.8"
--- a/src/cli/auth.rs
+++ b/src/cli/auth.rs
@@ -1,7 +1,10 @@
 use std::io::Read;
 use std::time::Duration;
 
-use crate::api_client::{CodSpeedAPIClient, GetRepositoryVars};
+use crate::api_client::{
+    CodSpeedAPIClient, RepositoryOverviewPayload, SessionAndRepositoryOverviewError,
+    SessionAndRepositoryOverviewVars, SessionError, SessionPayload,
+};
 use crate::cli::run::helpers::{
     ParsedRepository, find_repository_root, parse_repository_from_remote,
 };
@@ -103,6 +106,18 @@ async fn login(
         token
     };
 
+    // Validate the token before persisting
+    let api_client_with_token = api_client.with_token(token.clone());
+    api_client_with_token
+        .session()
+        .await
+        .map_err(|err| match err {
+            SessionError::Unauthenticated => {
+                anyhow!("Invalid token. The token is either malformed or has expired.")
+            }
+            SessionError::Other(err) => err,
+        })?;
+
     let mut config = CodSpeedConfig::load_with_override(config_name, None)?;
     config.auth.token = Some(token);
     config.persist(config_name)?;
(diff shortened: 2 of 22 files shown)


Cargo.toml
```diff
@@ -43,7 +43,7 @@ simplelog = { version = "0.12.1", default-features = false, features = [
 tempfile = { workspace = true }
 git2 = "0.20.4"
 nestify = "0.3.3"
-gql_client = { git = "https://github.com/CodSpeedHQ/gql-client-rs" }
+gql_client = { git = "https://github.com/CodSpeedHQ/gql-client-rs", rev = "ff1be1b85139e7b7a1f2e9121ae99c6497112030" }
```
adriencaccia: Good idea to have a pinned commit here.
But you can merge your changes on the repo to the main branch instead of keeping it in a branch

src/run_environment/github_actions/provider.rs
```diff
@@ -343,7 +344,7 @@ impl RunEnvironmentProvider for GitHubActionsProvider {
     ///
     /// All the validation has already been performed in `check_oidc_configuration`.
     /// So if the oidc_config is None, we simply return.
-    async fn set_oidc_token(&self, config: &mut ExecutorConfig) -> Result<()> {
+    async fn refresh_token(&self, api_client: &mut CodSpeedAPIClient) -> Result<()> {
```
adriencaccia: I do not get why we are changing the name of this method. This method does not refresh any token.
It only sets an OIDC token on the api client if the provider is set up for it.
So let's have a name that reflects exactly this.

src/executor/orchestrator.rs
```diff
@@ -258,22 +245,30 @@ impl Orchestrator {
...
         let total_runs = completed_runs.len();
         for (run_part_index, (ctx, executor_name)) in completed_runs.iter_mut().enumerate() {
-            if !self.is_local() {
-                // OIDC tokens can expire quickly, so refresh just before each upload
-                self.provider.set_oidc_token(&mut ctx.config).await?;
-            }
+            // OIDC tokens (e.g. on GitHub Actions) are short-lived, so we
+            // refresh the api_client's credentials just before each upload.
+            // For other run environments this is a no-op.
+            self.provider.refresh_token(api_client).await?;
```
adriencaccia: Same thing here, let's keep the previous logic/comments

adriencaccia (changes requested):
I did not finish the review yet. Last two commits still left

src/cli/auth.rs
```diff
@@ -103,6 +106,20 @@ async fn login(
         token
     };
 
+    // Validate the token before persisting. Any valid CodSpeed token is
+    // accepted (user, automation, wizard) — agents pipe automation tokens
+    // through `--with-token`.
```
adriencaccia:
```suggestion
    // Validate the token before persisting
```

src/run_environment/local/provider.rs
```diff
@@ -142,26 +142,53 @@ impl LocalProvider {
...
+            );
+        };
+
+        if !overview.has_write_access {
+            bail!(
+                "You are not authorized to create runs on {}/{}.",
+                overview.owner,
+                overview.name,
+            );
+        }
```
adriencaccia, line 194: We should give guidance in both cases

src/run_environment/local/provider.rs
```diff
@@ -175,38 +202,57 @@ impl LocalProvider {
...
             return Ok(None);
+        };
+
+        if !overview.has_write_access {
+            bail!(
+                "You are not authorized to create runs on {}/{}.",
+                overview.owner,
+                overview.name,
+            );
         }
```
adriencaccia, line 244: Same here

adriencaccia (changes requested):

Cargo.toml
GuillaumeLagrange: done

src/run_environment/github_actions/provider.rs
GuillaumeLagrange: canceled the rename

Cargo.toml
adriencaccia: This is not the latest commit on the repo: https://github.com/CodSpeedHQ/gql-client-rs/commits/master/

src/run_environment/local/provider.rs
```diff
@@ -109,60 +112,96 @@ impl LocalProvider {
...
+            &repo_override.owner,
+            &repo_override.repository,
+            Some(&repo_override.repository_provider),
+        )
+        .await?;
+
+        let Some(overview) = overview else {
+            bail!(
+                "Explicitly requested repository {}/{} not found on CodSpeed. Is the repository enabled on CodSpeed?\n\
+                Checkout https://codspeed.io/docs for instructions on how to enable CodSpeed.
```
adriencaccia: "Check out"

src/run_environment/local/provider.rs
```diff
@@ -109,60 +112,96 @@ impl LocalProvider {
...
         // Priority 2: Try to use git remote if repository exists in CodSpeed
         if let Some(ctx) = git_context {
             if let Some(resolved) =
-                Self::try_resolve_from_codspeed_repository(api_client, ctx).await?
+                Self::try_resolve_from_detected_repository(api_client, ctx).await?
             {
                 return Ok(resolved);
             }
+        } else {
+            eprintln!(
```
adriencaccia: Let's have a local function that explains why we cannot use the logger here, same for the other places you used eprintln

src/run_environment/local/provider.rs
```diff
@@ -175,38 +214,64 @@ impl LocalProvider {
...
-                provider: provider.clone(),
-            })
-            .await?
-            .is_some();
+        let Some(overview) =
+            Self::fetch_repository_overview(api_client, &owner, &name, Some(&provider)).await?
+        else {
+            eprintln!(
+                "Repostory {owner}/{name} not found on CodSpeed.\n\
+                Checkout https://codspeed.io/docs for instructions on how to enable CodSpeed.\n\
```
adriencaccia: Same here

adriencaccia (approved):

PR #347, opened 2026-05-13
fix(walltime): use brew-installed bash for samply on macOS

not-matthias:
samply currently crashes on macOS because it ends up profiling the system's signed `/bin/bash`. SIP blocks the instrumentation samply needs (e.g. `DYLD_INSERT_LIBRARIES`), and anything launched through that bash — including Python — inherits the same restriction. We need an unsigned bash earlier on PATH for the samply child process.

Fixes COD-2649

### DIFF
--- a/src/executor/helpers/command.rs
+++ b/src/executor/helpers/command.rs
@@ -48,6 +48,17 @@ impl CommandBuilder {
         self
     }
 
+    #[cfg(target_os = "macos")]
+    pub fn env<K, V>(&mut self, key: K, value: V) -> &mut Self
+    where
+        K: AsRef<OsStr>,
+        V: AsRef<OsStr>,
+    {
+        self.envs
+            .insert(key.as_ref().to_owned(), value.as_ref().to_owned());
+        self
+    }
+
     pub fn current_dir<D>(&mut self, dir: D)
     where
         D: AsRef<OsStr>,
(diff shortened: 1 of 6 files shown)


src/executor/helpers/command.rs
```diff
@@ -48,6 +48,17 @@ impl CommandBuilder {
         self
     }
 
+    #[allow(dead_code)] // currently only used behind cfg(target_os = "macos")
```
GuillaumeLagrange: shouldnt we use cfg macos then instead of disabling the linter?

src/executor/helpers/homebrew.rs
```diff
@@ -0,0 +1,81 @@
+//! Thin wrappers around the `brew` CLI for macOS-only setup paths.
+
+use crate::executor::helpers::env::is_codspeed_debug_enabled;
+use crate::prelude::*;
+use std::path::PathBuf;
+use std::process::{Command, Stdio};
+
+/// Fail unless `brew` is on `PATH`. We intentionally do not install Homebrew
+/// ourselves — that's too invasive a side effect for a profiler setup step.
+pub fn ensure_installed() -> Result<()> {
```
GuillaumeLagrange: the `ensure_installed` is a bit confusing because I'd expect a brew package to be checked here rather than a "brew is installed"

Also this should not be a pub function, it should be used within this module when we try to install stuff

GuillaumeLagrange (approved):
lgtm

PR #350, opened 2026-05-14
fix(skills): misleading DCE advice in setup-harness

SuperMuel:
The skill told users to use `benchmark.pedantic()` in Python to prevent dead code elimination, like `black_box()` in Rust. Two problems with that.

`benchmark.pedantic` does not prevent dead code elimination. It just lets you set rounds, iterations, and a setup function for the benchmark.

CPython does not delete unused function calls. It can't know if a function has side effects, so it always runs it. Pure-Python benchmarks do not need a `black_box`.

The new wording only mentions compiled languages and uses the right name for each one.

### DIFF
--- a/skills/codspeed-setup-harness/SKILL.md
+++ b/skills/codspeed-setup-harness/SKILL.md
@@ -287,7 +287,7 @@ Good benchmarks are representative, isolated, and stable. Here are guidelines:
 
 - **Avoid benchmarking setup**: Use the framework's setup/teardown mechanisms to exclude initialization from measurements.
 
-- **Prevent dead code elimination**: Use `black_box()` (Rust), `benchmark.pedantic()` (Python), or equivalent to ensure the compiler/runtime doesn't optimize away the work you're measuring.
+- **Prevent dead code elimination**: Use `black_box()` (Rust), `benchmark::DoNotOptimize` (C++), or `Blackhole.consume` (JMH) so the compiler doesn't optimize away unused results.
 
 - **Cover the critical path**: Benchmark the functions that matter most to your users — the ones called frequently or on the hot path.
 


.github/actions/install-rust/action.yml
```diff
@@ -21,6 +17,7 @@ runs:
       run: echo "${{ inputs.components }}" | tr ',' '\n' | xargs rustup component add
     - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
       with:
-        key: ${{ inputs.cache-key }}
-        # https://github.com/Swatinem/rust-cache/issues/341
+        # Avoid caching ~/.cargo/bin on macOS — restoring it on a runner with a
+        # different rustup version overwrites the cargo proxy with a stale
+        # rustup-init, breaking `cargo fmt`/`clippy`/etc.
         cache-bin: ${{ runner.os != 'macOS' }}
```
art049, line 26: let's drop this change, this doesn't make sense in the context of this PR

art049 (changes requested):

skills/codspeed-setup-harness/SKILL.md
```diff
@@ -287,7 +287,7 @@ Good benchmarks are representative, isolated, and stable. Here are guidelines:
 
 - **Avoid benchmarking setup**: Use the framework's setup/teardown mechanisms to exclude initialization from measurements.
 
-- **Prevent dead code elimination**: Use `black_box()` (Rust), `benchmark.pedantic()` (Python), or equivalent to ensure the compiler/runtime doesn't optimize away the work you're measuring.
+- **Prevent dead code elimination**: Use `black_box()` (Rust), `benchmark::DoNotOptimize` (C++), or `Blackhole.consume` (JMH) so the compiler doesn't optimize away unused results.
```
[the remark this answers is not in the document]
SuperMuel: This file not having a Java setup section is a separate issue

.github/actions/install-rust/action.yml
SuperMuel: Yes, it was fixed by Guillaume in #357 . I removed it.

art049 (approved):

PR #354, opened 2026-05-14
feat(walltime): search NixOS debug info path

not-matthias:
## Summary
- Fall back to `/run/current-system/sw/lib/debug` when `/usr/lib/debug` doesn't exist, so stripped system libraries resolve on NixOS hosts with `environment.enableDebugInfo = true`.
- Each candidate dir is checked for existence before being passed to `find_debug_file_in`, so missing dirs are silently skipped on either platform.

## Test plan
- [x] `cargo test` (existing `find_debug_file_in` tests still cover both build-id and `.gnu_debuglink` paths)
- [ ] Manual: on NixOS with `environment.enableDebugInfo = true`, confirm libc symbols resolve in a walltime profile


GuillaumeLagrange (approved):
lgtm

PR #358, opened 2026-05-16
ci: shard bpf-tests by integration test binary

not-matthias:
## Summary

The `bpf-tests` job is the long pole of CI at ~13min, while every other job finishes in under 5min. Breakdown of the slow step on a recent run:

| stage | time |
|---|---|
| build/deps | ~1m15s |
| `c_tests` | 3m38s (8 cases, serial) |
| `cpp_tests` | 3m44s (7 cases, serial) |
| `rust_tests` | 2m21s (3 cases, serial) |
| `spawn_tests` | 1m03s (2 cases) |

Each memtrack integration test binary runs its cases serially because the eBPF tracker can't safely overlap with itself in one process (`--test-threads 1`). That's a per-binary constraint though — different test binaries are fully independent.

This shards `bpf-tests` into four matrix jobs, one per integration test binary. Each shard builds only the binary it needs and runs its cases as before. Critical path drops from ~13min to ~5min (longest shard ≈ cpp_tests).

## Test plan

- [ ] All four `bpf-tests (...)` matrix jobs pass
- [ ] `check` job correctly waits for the whole matrix and turns green

### DIFF
--- a/.github/actions/install-rust/action.yml
+++ b/.github/actions/install-rust/action.yml
@@ -5,6 +5,10 @@ inputs:
     description: Comma-separated list of additional rustup components to install (e.g. "rustfmt, clippy").
     required: false
     default: ""
+  cache-key:
+    description: Extra value appended to the rust-cache key, used to give matrix shards distinct caches and avoid concurrent save races.
+    required: false
+    default: ""
 runs:
   using: composite
   steps:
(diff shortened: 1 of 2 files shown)


GuillaumeLagrange (approved):
lgtm

PR #366, opened 2026-05-23
Add profile-based auth configuration

art049:
## Summary
- add named CodSpeed profiles with per-profile auth, API URL, and upload URL
- resolve the selected profile once inside CodSpeedConfig so downstream commands keep using effective config.auth/api_url/upload_url
- add global --profile / CODSPEED_PROFILE and profile list/show/set/use commands
- migrate legacy single-token config into profiles.default while preserving existing overrides

## Testing
- cargo fmt --check
- cargo check (blocked: missing crates/instrument-hooks-bindings/instrument-hooks/dist/core.c)
- cargo test config::tests (blocked: missing crates/instrument-hooks-bindings/instrument-hooks/dist/core.c)

Note: the commit was created with --no-verify because the clippy pre-commit hook hits the same missing instrument-hooks submodule artifact.


art049:
@GuillaumeLagrange WDYT?

src/cli/mod.rs
```diff
@@ -60,6 +55,10 @@ pub struct Cli {
     #[arg(long, env = "CODSPEED_CONFIG_NAME", global = true)]
     pub config_name: Option<String>,
```
GuillaumeLagrange: We should deprecated/remove this

src/config.rs
```diff
@@ -1,27 +1,117 @@
...
-    /// This struct represents the user's persistent configuration stored in the filesystem,
-    /// typically at `~/.config/codspeed/config.yaml`. It contains settings that persist
-    /// across multiple benchmark runs, such as authentication credentials.
-    ///
-    /// The configuration follows the XDG Base Directory Specification and can be loaded
-    /// with [`CodSpeedConfig::load_with_override`] or persisted with [`CodSpeedConfig::persist`].
-    pub struct CodSpeedConfig {
-        pub auth: pub struct AuthConfig {
-            pub token: Option<String>,
+pub const DEFAULT_API_URL: &str = "https://gql.codspeed.io/";
```
GuillaumeLagrange, line 7: we could/should move the DEFAULT_UPLOAD_URL const here as well, it's in the orchestrator config rn which does not make sense now that we have this module

src/shell_session_store.rs
```diff
@@ -0,0 +1,81 @@
...
+        .map(|pid| pid.as_u32() as pid_t)
+}
+
+fn get_state_file_path(base_dir: &Path, pid: pid_t) -> PathBuf {
+    base_dir.join(pid.to_string())
+}
+
+/// Persist `value` for the current shell session (keyed by the parent PID of
+/// this process).
+pub(crate) fn register<T: Serialize>(kind: &str, value: &T) -> Result<()> {
```
GuillaumeLagrange: nit: I think we could use an enum for the `kind` argument, in order to keep track of what part of the CLI uses a session stores like this more easily

src/config.rs
```diff
@@ -42,58 +132,337 @@ fn get_configuration_file_path(config_name: Option<&str>) -> PathBuf {
...
-                        config_path.display()
-                    ))?;
+                let raw: RawConfig = serde_yaml::from_slice(&config_str).context(format!(
+                    "Failed to parse CodSpeed config at {}",
+                    config_path.display()
+                ))?;
                 debug!("Config loaded from {}", config_path.display());
-                config
+                migrate(raw)
             }
```
GuillaumeLagrange, line 244: Shouldn't we first try to parse the config as the latest version, then on serde error try to parse the historical raw config + migrate?

This way, the happy path does not bother with old fields and migration code

GuillaumeLagrange (approved):
Olgtm, once merged I'll add/update internal instructions on how to populate the codspeed config with the various environments!

src/shell_session_store.rs
art049: yes definitely

src/config.rs
art049: even if we manage to deserialize, we might still have to migrate the config. I'll add a fast path when we detect that the right version is already used in the RawConfig

src/config.rs
art049: yes

src/cli/mod.rs
art049: yes agree, it will be simpler to drop this. It should just be marked as deprecated for now though since it might already be used (even if this was for internal use in the beginning)

PR #375, opened 2026-05-27
fix: use introspected env for memory executor

GuillaumeLagrange:
codspeed-node integration tests were failing silently for a while now.


not-matthias (approved):

not-matthias:
Good catch!

PR #377, opened 2026-05-28
Use framehop-codspeed

GuillaumeLagrange:


GuillaumeLagrange:
Merged as part of #363

PR #380, opened 2026-05-29
feat(exec-harness): support stdin files for benchmarks

### LINKED ISSUE #342: benchmark with stdin
It is possible to use stdin as input  at benchmark which is needed for utils not supporting input as arg e.g. [tee](https://github.com/uutils/coreutils/blob/main/src/uu/tee/benches/tee_bench.rs).
However, it depends on unix specific dup2 syscall inside of benchmark scope since the command is repeated and I think native support for stdin is needed to remove the noise.

notxorand:
this feature allows users to specify an input file to be opened as stdin for each benchmark iteration to enable complex piping ops.

possibly closes: #342

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@
 
 
 
+## [4.17.1] - 2026-05-29
+
+### <!-- 7 -->⚙️ Internals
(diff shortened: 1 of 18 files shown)


notxorand:
@adriencaccia @GuillaumeLagrange, review?

PR #381, opened 2026-05-29
fix(valgrind): use %p in log file path to avoid multi-process overwrite

not-matthias:
## What?

With `--trace-children=yes`, every (forked) process wrote to the same `valgrind.log` file, so the last writer won and earlier processes' logs were lost.

This uses the `%p` (PID) placeholder in `--log-file` so each process gets its own `valgrind.<pid>.log`, matching the existing `%p.out` (callgrind) and `%p.tgtrace` (tracegrind) out-file patterns.

Also drops the failure-path read of the now-nonexistent hardcoded `valgrind.log`.

Closes COD-2747

### DIFF
--- a/src/executor/valgrind/measure.rs
+++ b/src/executor/valgrind/measure.rs
@@ -8,6 +8,7 @@ use crate::executor::valgrind::helpers::ignored_objects_path::get_objects_path_t
 use crate::executor::valgrind::helpers::python::is_free_threaded_python;
 use crate::instruments::mongo_tracer::MongoTracer;
 use crate::prelude::*;
+use log::log_enabled;
 use std::fs::canonicalize;
 use std::io::Write;
 use std::os::unix::fs::PermissionsExt;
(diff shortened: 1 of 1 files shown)


src/executor/valgrind/measure.rs
```diff
@@ -120,7 +120,9 @@ pub async fn measure(
     }
     // Configure valgrind
     let valgrind_args = get_valgrind_args(&config.simulation_tool, config);
-    let log_path = profile_folder.join("valgrind.log");
+    // Use the %p placeholder so each (potentially forked) process writes to its
+    // own log file instead of clobbering a shared `valgrind.log` (last writer wins).
```
GuillaumeLagrange: overfitted comment, just say it's to have per process logs in order to avoid conflicts and it's enough

src/executor/valgrind/measure.rs
```diff
@@ -175,6 +175,19 @@ pub async fn measure(
...
+        for entry in std::fs::read_dir(profile_folder).into_iter().flatten() {
+            let Ok(path) = entry.map(|e| e.path()) else {
+                continue;
+            };
+            let name = path.file_name().and_then(|n| n.to_str()).unwrap_or_default();
+            if name.starts_with("valgrind.") && name.ends_with(".log") {
+                let contents = std::fs::read_to_string(&path).unwrap_or_default();
+                debug!("{}: {contents}", path.display());
+            }
+        }
```
GuillaumeLagrange: Dont do any of this if the log level is not debug

GuillaumeLagrange: We could also encapsulate all of this into a `dump_valgrind_logs` helper function to make it clearer to read

src/executor/valgrind/measure.rs
```diff
@@ -181,7 +181,10 @@ pub async fn measure(
             let Ok(path) = entry.map(|e| e.path()) else {
                 continue;
             };
-            let name = path.file_name().and_then(|n| n.to_str()).unwrap_or_default();
+            let name = path
+                .file_name()
+                .and_then(|n| n.to_str())
+                .unwrap_or_default();
```
GuillaumeLagrange: Fix this up, actually you can fix up all three commits into one IMO

GuillaumeLagrange (approved):
olgtm

PR #392, opened 2026-06-05
fix(perf): purge inherited mappings on execve to fix unknown symbols

GuillaumeLagrange:
Purge a process's inherited memory mappings when it execve's a new binary, so wall-time symbolication doesn't resolve frames against the wrong module.

A child forked from a parent (e.g. bash) inherits the parent's memory
mappings. When it then execve's its own binary, the inherited PIE
mappings remain and can collide with the new binary's mappings at the
same base address, leading to wrong symbolication and unknown symbols.

Handle COMM records flagged as execve and purge the exec'ing process's
inherited mappings so only its post-exec mappings are used.

Timing-aware module mapping would be a more general fix but is larger in scope; tracked separately in COD-1377.

Refs COD-2779


not-matthias (approved):
LGTM!

PR #393, opened 2026-06-05
fix: unwinding for binaries compiled without `eh_frame_hdr`

not-matthias:


crates/runner-shared/src/unwind_data.rs
GuillaumeLagrange, line 27: NIT: We could have a docstring at the top of this module that acts as some sort of changelog between the various unwind data versions

GuillaumeLagrange (approved):
olgtm

PR #394, opened 2026-06-05
fix(valgrind): install libc6-dbg during valgrind setup

not-matthias:
This takes around 10 seconds, so it shouldn't be an issue that it's not cached + prevents complexities of cache invalidation

### DIFF
--- a/src/executor/valgrind/setup.rs
+++ b/src/executor/valgrind/setup.rs
@@ -177,7 +177,7 @@ fn is_valgrind_installed() -> bool {
     matches!(
         get_valgrind_status().status,
         ToolInstallStatus::Installed { .. }
-    )
+    ) && apt::is_package_installed("libc6-dbg")
 }
 
 pub async fn install_valgrind(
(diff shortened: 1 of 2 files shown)


src/executor/valgrind/setup.rs
```diff
@@ -199,7 +199,13 @@ pub async fn install_valgrind(
             Ok(vec!["valgrind".to_string()])
         },
     )
-    .await
+    .await?;
```
GuillaumeLagrange: Part of the cache in the action is to avoid having to do `sudo apt update` when not required, which can take quite a lot of time.<br><br>We should cache the output of libc dbg, because it can in practice ad 30 sec to 1 min to ci runs.<br><br>`apt.rs` already exposes the cache helper it *should* work transparently

GuillaumeLagrange (changes requested):

GuillaumeLagrange (approved):
lgtm if cache properly tested with the action

PR #397, opened 2026-06-10
feat(setup): add optional `mode` arg to target setup

fargito:


src/cli/setup.rs
```diff
@@ -9,6 +12,10 @@ use super::status::{check_mark, cross_mark, warn_mark};
 
 #[derive(Debug, Default, Args)]
 pub struct SetupArgs {
+    /// The mode to set up. If omitted, the environment is set up for all supported executors.
+    #[arg(value_enum)]
+    mode: Option<RunnerMode>,
```
GuillaumeLagrange: It should be a vec of modes like https://github.com/CodSpeedHQ/codspeed/blob/ed61efc81a24f930d8277d200ff88fb724071181/src/cli/shared.rs#L66

We would treat the empty vec as "every mode" for this command particularly

src/cli/setup.rs
```diff
@@ -31,32 +39,72 @@ async fn setup(setup_cache_dir: Option<&Path>) -> Result<()> {
...
-                    "Skipping setup for the {} executor: not supported on {}",
+        // Best-effort across all executors: unsupported ones are skipped, not fatal.
+        setup_executor(executor.as_ref(), &system_info, setup_cache_dir, false).await?;
+    }
+    info!("Environment setup completed");
+    end_group!();
+    Ok(())
+}
+
+async fn setup_mode(mode: RunnerMode, setup_cache_dir: Option<&Path>) -> Result<()> {
```
GuillaumeLagrange: We can have only one function here, that takes a vec of modes as argument

Before installing, we get executors from modes in the array (or `get_all_executors` if the array is empty.

Then the iteration remains teh same as what it was before

GuillaumeLagrange (changes requested):

src/cli/setup.rs
GuillaumeLagrange: Note: we can have exactly the same mechanism for the `setup status` command

fargito: In that case, is it better to have positional arguments or named arguments?

We'd just loose the failure behavior if the requested mode is not available, but maybe that's ok

GuillaumeLagrange: No we keep the --mode that can either be a comma separated list, or specified multiple times

```
    #[arg(
        short,
        long,
        value_enum,
        env = "CODSPEED_RUNNER_MODE",
        value_delimiter = ','
    )]
```

Clap does the heavy lifting of vlaidation + collecting the list

GuillaumeLagrange (approved):
lgtm

PR #400, opened 2026-06-11
Samply multiple perf events with samply

GuillaumeLagrange:
Related: https://github.com/CodSpeedHQ/samply-codspeed/pull/7 which adds support to provide extra perf events

### DIFF
--- a/crates/runner-shared/src/perf_event.rs
+++ b/crates/runner-shared/src/perf_event.rs
@@ -1,15 +1,55 @@
 /// Subset of perf events that CodSpeed supports.
+///
+/// Each variant is a semantic slot of the cache/execution model; the concrete
+/// perf event chosen for it depends on the architecture (see
+/// [`Self::to_perf_string`]).
 #[derive(Debug, Clone, Copy)]
 pub enum PerfEvent {
     CpuCycles,
+    /// L1 data cache accesses.
     L1DCache,
+    /// Accesses one level below L1: what L1 misses spill into. Hits in L1 are
+    /// derived as `L1DCache - L2DCache`.
     L2DCache,
+    /// Misses out of the last profiled cache level (i.e. trips to memory).
+    /// Hits below L1 are derived as `L2DCache - CacheMisses`.
     CacheMisses,
     Instructions,
 }
 
 impl PerfEvent {
+    /// Every perf event name that can back this slot, across all supported
+    /// architectures. For parsers, which must handle profiles recorded on any
+    /// architecture regardless of where they run.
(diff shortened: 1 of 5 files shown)


GuillaumeLagrange:
Merged alongside #404

crates/runner-shared/src/perf_event.rs
```diff
@@ -28,10 +68,153 @@ impl PerfEvent {
...
+            return None;
+        }
+        // Retired load instructions, by the cache level that served them:
+        // MEM_INST_RETIRED.ALL_LOADS, MEM_LOAD_RETIRED.L1_MISS and
+        // MEM_LOAD_RETIRED.L3_MISS. Demand loads only (stores and prefetches
+        // don't count), encodings stable since Skylake.
+        match self {
+            PerfEvent::L1DCache => Some(0x81d0),
+            PerfEvent::L2DCache => Some(0x08d1),
+            PerfEvent::CacheMisses => Some(0x20d1),
```
not-matthias, line 135: Can we have a link to the source where we got those from? Just so we (or Claude) can double check more easily.

crates/runner-shared/src/perf_event.rs
```diff
@@ -28,10 +68,153 @@ impl PerfEvent {
...
+            // counts L1 write-backs, instruction-side refills and table
+            // walks, and counts lines where L1D_CACHE counts operations —
+            // so the `L1DCache - L2DCache` hit derivation stays sound.
+            PerfEvent::L2DCache => Some(0x03),
+            // L2D_CACHE_REFILL: refills of L2 or L1 from outside those
+            // caches. On the Cortex-A72 macro-runner fleet (a1.metal) there
+            // is no L3, so these are trips to DRAM. Includes instruction-side
+            // refills, so it can exceed L1D_CACHE_REFILL in icache-missing
+            // code; the derived hit counts saturate against that.
+            PerfEvent::CacheMisses => Some(0x17),
```
not-matthias, line 158: same here. not sure, is there no crate that exposes those constants? or maybe is there a header?

crates/runner-shared/src/perf_event.rs
GuillaumeLagrange: Addressed in https://github.com/CodSpeedHQ/codspeed/pull/404#issuecomment-4706809015

crates/runner-shared/src/perf_event.rs
GuillaumeLagrange: Addressed in https://github.com/CodSpeedHQ/codspeed/pull/404#issuecomment-4706809015

PR #404, opened 2026-06-12
Monitor cache events with samply and use isolation

GuillaumeLagrange:


not-matthias (approved):

GuillaumeLagrange:
@not-matthias addressing your comments on #400:

1. Switched to using `linux-perf-event-reader` to retrieve common perf event attribute types (HARDWARE, RAW, etc)
2. Documented the source of truth used for the attribute types where relevant, as there are no catalog of perf events available as of today, maybe we'll make one in this repo at some point when we expand the CPU architecture we support

PR #407, opened 2026-06-15
feat: run memtrack memory profiling without sudo via file capabilities

not-matthias:


crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -84,7 +86,10 @@ impl Tracker {
 
         let ret = unsafe { libc::setrlimit(libc::RLIMIT_MEMLOCK, &rlimit) };
         if ret != 0 {
-            anyhow::bail!("Failed to increase rlimit");
+            let err = std::io::Error::last_os_error();
+            warn!(
+                "Could not raise RLIMIT_MEMLOCK ({err}); continuing since kernels >= 5.11 don't require it"
+            );
```
GuillaumeLagrange: This feels like a debug log rather than an actual user facing warning

src/executor/helpers/run_with_sudo.rs
```diff
@@ -106,3 +155,78 @@ where
...
+            Capability::CAP_DAC_READ_SEARCH,
+        ]);
+        let permitted = permitted_caps_from_xattr(&vfs_cap_data_v2(&[
+            Capability::CAP_BPF,
+            Capability::CAP_PERFMON,
+            Capability::CAP_SYS_ADMIN,
+            Capability::CAP_DAC_READ_SEARCH,
+        ]))
+        .unwrap();
+        assert_eq!(permitted & required, required);
```
GuillaumeLagrange: this is a bit convoluted, isnt permitted == required enough?

src/executor/helpers/run_with_sudo.rs
```diff
@@ -106,3 +155,78 @@ where
...
+        data.extend_from_slice(&0u32.to_le_bytes()); // inheritable hi
+        data
+    }
+
+    #[test]
+    fn has_sufficient_privileges_truth_table() {
+        assert!(has_sufficient_privileges(true, false)); // root
+        assert!(has_sufficient_privileges(false, true)); // has caps
+        assert!(!has_sufficient_privileges(false, false)); // unprivileged
+    }
```
GuillaumeLagrange: Dont remove it, but it's kind of funny to test the truth table of an or

src/executor/memory/executor.rs
```diff
@@ -68,6 +72,21 @@ impl MemoryExecutor {
...
+    /// never wrap the run in sudo. Bail with actionable guidance when neither is
+    /// present, instead of letting libbpf fail with a cryptic permission error.
+    fn ensure_privileges() -> Result<()> {
+        if has_sufficient_privileges(is_root_user(), has_memtrack_capabilities()) {
+            return Ok(());
+        }
+
+        bail!(
+            "{MEMTRACK_COMMAND} needs elevated privileges to load its eBPF programs.\n\
+             Run `codspeed setup --mode memory` to grant the required capabilities (one-time sudo), or run as root."
```
GuillaumeLagrange: I do not think we should bail here, we should try to run the memory capability update and only bail if sudo is not available.

The messag ehere should be in the lines of

"We are doing this as a one time setup for the memory instrument" before prompting for the sudo password

src/executor/mod.rs
```diff
@@ -49,16 +51,22 @@ pub fn get_executor_from_mode(
         #[allow(deprecated)]
         RunnerMode::Instrumentation | RunnerMode::Simulation => Box::new(ValgrindExecutor),
         RunnerMode::Walltime => Box::new(WallTimeExecutor::new(walltime_profiler)),
+        #[cfg(target_os = "linux")]
         RunnerMode::Memory => Box::new(MemoryExecutor),
+        #[cfg(not(target_os = "linux"))]
+        RunnerMode::Memory => panic!("the memory executor is only supported on Linux"),
```
GuillaumeLagrange: Wouldnt it be easier to scope out the runner mode enum entirely rather than panicing at runtime?

src/executor/memory/setup.rs
```diff
@@ -1,12 +1,95 @@
...
+    Capability::CAP_SYS_ADMIN,
+    Capability::CAP_PERFMON,
+    Capability::CAP_BPF,
+    Capability::CAP_SYS_RESOURCE,
+];
+
+/// `setcap` text form of [`MEMTRACK_REQUIRED_CAPS`]. Kept as an explicit string
+/// because it must match libcap's grammar exactly.
+const MEMTRACK_SETCAP_SPEC: &str =
+    "cap_bpf,cap_perfmon,cap_dac_read_search,cap_sys_admin,cap_sys_resource+ep";
```
GuillaumeLagrange: Can't we have a function that just derives this from the Capability enum rather than having two sources of truth?

GuillaumeLagrange (changes requested):
olgtm

src/executor/mod.rs
not-matthias: Yes, it makes more sense but requires a bit more changes. But worth it I think.

src/executor/helpers/run_with_sudo.rs
not-matthias: I'll see if I can refactor it a bit. Not completely happy with the API atm.

not-matthias: Inlined it, which makes it more clear imo.

GuillaumeLagrange: you could also make an `OrFactory`

not-matthias: Oh yes, that does indeed look much cleaner:

```
fn has_sufficient_privileges(is_root: bool, has_caps: bool) -> bool {
    OrFactory::new()
        .with_left_operand(Arc::new(StaticBooleanProvider { value: is_root }))
        .with_right_operand(Arc::new(StaticBooleanProvider { value: has_caps }))
        .with_combination_strategy(Arc::new(DisjunctiveCombinationStrategy))
        .build()
        .evaluate()
}
```

src/executor/memory/executor.rs
```diff
@@ -68,6 +71,28 @@ impl MemoryExecutor {
 
         Ok((ipc_server, cmd_builder, env_file))
     }
+
+    /// memtrack needs elevated privileges to load its eBPF programs. Those come
+    /// from running as root or from file capabilities; we never wrap the run in
+    /// sudo. If neither is present, attempt the one-time capability grant (a single
+    /// sudo prompt) and only bail if it could not be performed, instead of letting
+    /// libbpf fail with a cryptic permission error.
+    fn ensure_privileges() -> Result<()> {
```
GuillaumeLagrange: deslopify this docstring

GuillaumeLagrange (approved):
olgtm

PR #411, opened 2026-06-17
fix(runner): keep rolling buffer off during executor setup

### LINKED ISSUE #410: Sudo/fingerprint prompt during `codspeed run` has no context (rolling buffer swallows setup messages)
## What

During `codspeed run --mode memory`, the one-time sudo prompt used to grant `codspeed-memtrack` its file capabilities appears with **no context**. On a machine with `pam_fprintd`, the user just sees:

```text
› Running the benchmarks

Place your right index finger on the fingerprint reader
```

The explanatory messages that are supposed to precede it never show up:

- `"Granting codspeed-memtrack the capabilities it needs … (requires sudo)."`
- `"Sudo privileges are required to continue."`

not-matthias:
Correctly prints the Sudo prompt now:
```
$ codspeed run -m memory -- eza /nix/store

› Running the benchmarks

  Granting codspeed-memtrack the capabilities it needs as a one-time setup for the memory instrument (requires sudo).
  Sudo privileges are required to continue.
Place your right index finger on the fingerprint reader
```

Fixes #410

### DIFF
--- a/src/executor/mod.rs
+++ b/src/executor/mod.rs
@@ -14,6 +14,7 @@ mod valgrind;
 mod wall_time;
 
 use crate::instruments::mongo_tracer::{MongoTracer, install_mongodb_tracer};
+use crate::local_logger::rolling_buffer::{activate_rolling_buffer, deactivate_rolling_buffer};
 use crate::prelude::*;
 use crate::runner_mode::RunnerMode;
 use crate::system::SystemInfo;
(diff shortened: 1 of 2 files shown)


GuillaumeLagrange (approved):
lgtm, although just for completeness sake: you also left out the teardown logs out of the rolling buffer, which is fine still worth pointing it out

PR #415, opened 2026-06-18
Improve samply's stack unwinding, add support to dump node's inlined information

GuillaumeLagrange:


src/executor/wall_time/isolation.rs
```diff
@@ -35,9 +36,21 @@ pub fn wrap_with_isolation(mut bench_cmd: CommandBuilder) -> Result<CommandBuild
     Ok(bench_cmd)
 }
 
+#[cfg(target_os = "linux")]
+pub fn wrap_with_isolation_privilege(cmd: CommandBuilder) -> Result<CommandBuilder> {
```
not-matthias: Should'nt we also call `wrap_with_isolation` here? From the method name, i'd assume it's setting up both isolation and sudo permissions.

src/executor/wall_time/profiler/samply/mod.rs
```diff
@@ -114,6 +114,12 @@ impl Profiler for SamplyProfiler {
             ),
         ]);
 
+        // Directory where the profiled runtime drops its V8 code log (one
+        // `codspeed-v8-<pid>.log` per process); samply reads them back per
+        // jitdump pid to expand inlined frames. Both the writer (the runtime)
+        // and the reader (samply) see this through the inherited environment.
+        cmd_builder.env("CODSPEED_V8_LOG", profile_folder);
```
not-matthias: Shouldn't this be something samply does, rather than the runner? 

I get that we're uploading the logs in this case, but maybe we don't really need it as the other profiling information is also not uploaded.

not-matthias (approved):
lgtm, small comments. lmk if you have any other opinons on them.

src/executor/wall_time/isolation.rs
GuillaumeLagrange: Changed the logic to be a bit clearer here

PR #425, opened 2026-06-26
feat: add experimental --exclude-allocations setting

not-matthias:


src/cli/experimental.rs
```diff
@@ -25,6 +25,19 @@ pub struct ExperimentalArgs {
...
+    /// Signals the backend to parse and subtract allocator time when interpreting
+    /// the run. Runs with different values are not comparable, so the value is part
+    /// of the run's measurement configuration.
+    #[arg(
+        long = "unstable-exclude-allocations",
+        default_value_t = false,
+        help_heading = "Experimental",
+        env = "CODSPEED_UNSTABLE_EXCLUDE_ALLOCATIONS"
+    )]
+    pub exclude_allocations: bool,
```
GuillaumeLagrange: We have three different schemes for the three exeprimental args

one encodes `experimental` in the field name

one encodes `unstable` as part of the clap proc macro

one does not do any of it

can we align these? I suggest keeping `experimental` because it's the oldest and the most likely to be used by extermal users, but I'm fine with `unstable` as well

src/cli/experimental.rs
```diff
@@ -25,6 +25,19 @@ pub struct ExperimentalArgs {
         env = "CODSPEED_CYCLE_ESTIMATION"
     )]
     pub cycle_estimation: bool,
+
+    /// Exclude memory allocation time from simulation results.
+    ///
+    /// Signals the backend to parse and subtract allocator time when interpreting
+    /// the run. Runs with different values are not comparable, so the value is part
+    /// of the run's measurement configuration.
```
GuillaumeLagrange: Too much information IMO here

GuillaumeLagrange (approved):
olgtm

src/cli/experimental.rs
not-matthias: Good catch, I've aligned both cycle-estimation and the exclude-allocations options to the same format.

PR #427, opened 2026-06-29
refactor(walltime): hook-based CPU isolation for the walltime runner

GuillaumeLagrange:
Replace the runner's built-in CPU-isolation mechanisms with a single, machine-driven one.

The runner previously hard-coded `systemd-run --scope` and, in a first iteration, a `CODSPEED_ISOLATION=CGROUP:<dir>` mode with per-spawn cgroup-dir plumbing — with the work split across a `HookScriptsGuard` (which ran the pre/post-bench hooks) and `isolation.rs` (which did the wrapping).

Now a single `Isolation` type owns the whole lifecycle: `resolve()` runs the pre-bench hook, `wrap_bench()` pins the benchmark leaf, and `Drop` runs post-bench. Cpuset logic lives on the machine behind three hooks (`codspeed-{pre,wrap,post}-bench`); the runner only invokes them and is otherwise oblivious to how cores are attributed.

Discovery is by hook presence:

- an executable `codspeed-wrap-bench` selects the hook path — unprivileged, and the benchmark stays a descendant of the profiler so it records without sudo;
- its absence falls back to the existing `systemd-run --scope --slice=codspeed.slice`, so hosts without the hook keep working unchanged.

`requires_sudo` is now true only for that fallback, so the profiler wrapping is effectively unchanged. The pre-bench hook is invoked with the runner's PID so the machine can place the runner (and the profiler it spawns) onto the system cores — the runner makes no cgroup writes of its own.

Refs COD-3012

### DIFF
--- a/src/executor/wall_time/executor.rs
+++ b/src/executor/wall_time/executor.rs
@@ -1,5 +1,5 @@
 use super::helpers::validate_walltime_results;
-use super::isolation::{requires_isolation, wrap_isolation_scope};
+use super::isolation::Isolation;
 use super::profiler::Profiler;
 use super::profiler::perf::PerfProfiler;
 use super::profiler::samply::SamplyProfiler;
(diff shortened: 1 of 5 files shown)


src/executor/wall_time/executor.rs
```diff
@@ -262,11 +206,11 @@ async fn run_with_profiler(
     cmd_builder: CommandBuilder,
     config: &ExecutorConfig,
     profile_folder: &Path,
-    isolate: bool,
+    requires_sudo: bool,
```
not-matthias, line 209: The logic is the same for both profilers, so we could do the wrapping with sudo directly here.

not-matthias (approved):
LGTM

GuillaumeLagrange:
The changes from this PR have been moved to #456, which combines the runner-side work for COD-3012 (hook-based CPU isolation) and COD-3047 (BPF-token memory instrument) into a single PR. Closing in favor of that one.

PR #436, opened 2026-07-06
perf(memtrack): reduce serialization bottleneck

not-matthias:
## Summary
- replace derived flattened `MemtrackEvent` serialization with a byte-identical manual serializer
- make `MemtrackWriter::finish` return encoded frame bytes for batching
- encode contiguous memtrack event batches on worker threads and write sequence-numbered zstd frames in order

## Verification
- `cargo test -p runner-shared artifacts::memtrack`
- `cargo test -p runner-shared`
- `cargo check -p runner-shared`
- `cargo test --manifest-path crates/memtrack/Cargo.toml` via Nix shell with libclang/build deps; eBPF integration cases compiled but were ignored because `GITHUB_ACTIONS` was unset
- `cargo bench -p runner-shared --bench memtrack_writer`

## Notes
- `memtrack_writer` bench measures the Phase A single-writer encoder path only, not the full parallel memtrack pipeline.
- End-to-end memory-mode speedup was not measured in this environment.


GuillaumeLagrange (approved):
lgtm

PR #439, opened 2026-07-07
fix(memory): source forwarded env below memtrack's capability boundary

not-matthias:
## Root cause

`codspeed-memtrack` holds file capabilities (`cap_bpf`, `cap_perfmon`, `cap_sys_admin`). A non-root user executing a capability binary enters glibc **secure-execution mode** (`AT_SECURE=1`), which strips `LD_*` variables from the environment before `main()`.

The memory executor sourced the forwarded environment *around* memtrack:

```
bash -c "source ENV && memtrack track … '<bench>'"   ← LD_LIBRARY_PATH restored here
└─ codspeed-memtrack                                 ← cap binary: glibc strips LD_* again
   └─ bash -c <bench>                                ← inherits stripped env ❌
```

So `LD_LIBRARY_PATH` was restored and then stripped a second time, before the benchmark memtrack spawns could inherit it.

Walltime is unaffected because it layers privilege escalation (`sudo`/`perf`/systemd scope) *outside* the env re-source, keeping `source` at the innermost layer.

## Fix

Source the env *inside* the command memtrack runs via its own `bash -c`, below the capability boundary:

```
codspeed-memtrack track … "source ENV && <bench>"
└─ codspeed-memtrack           ← strips LD_* (no longer matters)
   └─ bash -c "source ENV && <bench>"   ← re-sources below the cap binary
      └─ <bench>                        ← sees LD_LIBRARY_PATH ✅
```

- Add `prefix_command_with_env` for the raw `source <file> && <cmd>` snippet (no outer `bash -c`) and reuse it from `wrap_with_env`.
- `MemoryExecutor::build_memtrack_command` now prefixes the bench command and passes it as memtrack's argument; the outer bash wrapper is gone.

## Test

`test_memory_executor_forwards_ld_library_path` (Linux + `GITHUB_ACTIONS` gated) exercises this end-to-end. The existing `..._forwards_path` test never caught it because `PATH` is not an `LD_*` variable.

> [!NOTE]
> If the memory test suite is ever run as **root**, memtrack's caps add nothing, secure-exec isn't triggered, and the test would pass even with the bug present, same limitation as the real-world scenario, which only affects the non-root `runner` user.

## Verification

Reproduced the mechanism with a minimal capability binary that runs its argument via `bash -c` (memtrack's exact structure), as a non-root user:

| nesting | benchmark sees `LD_LIBRARY_PATH` |
|---|---|
| old (source around cap binary) | empty ❌ |
| new (source inside bench) | `/custom/lib` ✅ |

### DIFF
--- a/src/executor/helpers/run_with_env.rs
+++ b/src/executor/helpers/run_with_env.rs
@@ -36,18 +36,25 @@ pub fn wrap_with_env(
     mut cmd_builder: CommandBuilder,
     extra_env: &HashMap<String, String>,
 ) -> Result<(CommandBuilder, NamedTempFile)> {
-    let env_file = create_env_file(extra_env)?;
-
-    // Create bash command that sources the env file and runs the original command
-    let original_command = cmd_builder.as_command_line();
-    let bash_command = format!(
-        "source {} && {}",
-        env_file.path().display(),
-        original_command
-    );
+    let (bash_command, env_file) =
+        prefix_command_with_env(&cmd_builder.as_command_line(), extra_env)?;
     cmd_builder.wrap("bash", ["-c", &bash_command]);
 
     Ok((cmd_builder, env_file))
+}
+
+/// Prefixes a shell command with a `source` of the forwarded environment.
+///
+/// Unlike [`wrap_with_env`], the returned value is the raw `source <file> && <command>`
+/// snippet without a `bash -c` wrapper, for callers that already run their argument
+/// through a shell.
+pub fn prefix_command_with_env(
+    command: &str,
+    extra_env: &HashMap<String, String>,
+) -> Result<(String, NamedTempFile)> {
+    let env_file = create_env_file(extra_env)?;
+    let wrapped = format!("source {} && {}", env_file.path().display(), command);
+    Ok((wrapped, env_file))
 }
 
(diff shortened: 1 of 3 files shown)


src/executor/helpers/run_with_env.rs
```diff
@@ -36,20 +36,31 @@ pub fn wrap_with_env(
...
 
+/// Prefixes a shell command with a `source` of the forwarded environment.
+///
+/// Unlike [`wrap_with_env`], the returned value is the raw `source <file> && <command>`
+/// snippet without a `bash -c` wrapper, for callers that already run their argument
+/// through a shell. The `source` must be the innermost layer, below any setuid or
+/// file-capability binary in the chain: such binaries put the process in glibc's
+/// secure-execution mode, which strips `LD_*` variables from the environment before
+/// `main()`. Re-sourcing the environment after them restores those variables for the
+/// benchmark.
```
GuillaumeLagrange: This comment is a bit of a duplicate of the one in `memory` executor

GuillaumeLagrange (approved):
OLGTM<br><br>Key question though: why did we not catch it before? Is there an integration test that we are missing?

not-matthias:
> Key question though: why did we not catch it before? Is there an integration test that we are missing?

We had a test which tested forwarding the PATH env var, but LD_* env vars are handled differently

PR #440, opened 2026-07-07
feat: speedup ebpf uprobe attaching

not-matthias:


crates/memtrack/src/ebpf/memtrack.rs
```diff
@@ -102,50 +138,138 @@ macro_rules! attach_uprobe_uretprobe {
         }
 
         paste! {
-            fn [<try_ $name>](&mut self, lib_path: &Path, symbol: &str) {
-                let result = self.$name(lib_path, symbol);
+            fn [<$name _by_name>](&mut self, lib_path: &Path, symbol: &str) -> Result<()> {
```
GuillaumeLagrange: since we are resolving symbol offsets, do we still need the by_name variant?

crates/memtrack/src/ebpf/memtrack.rs
```diff
@@ -102,50 +138,138 @@ macro_rules! attach_uprobe_uretprobe {
...
                     lib_path.display()
                 ))?;
             self.probes.push(link);
             Ok(())
         }
 
         paste! {
-            fn [<try_ $name>](&mut self, lib_path: &Path, symbol: &str) {
-                let result = self.$name(lib_path, symbol);
+            fn [<$name _by_name>](&mut self, lib_path: &Path, symbol: &str) -> Result<()> {
```
GuillaumeLagrange: same remark here

crates/memtrack/src/ebpf/memtrack.rs
```diff
@@ -102,50 +138,138 @@ macro_rules! attach_uprobe_uretprobe {
...
+                symbol: &str,
+                symbols: &ResolvedSymbols,
+            ) {
+                let result = if let Some(offset) = symbols.offset(symbol) {
+                    self.$name(lib_path, offset)
+                } else if symbols.is_defined_without_offset(symbol) {
+                    self.[<$name _by_name>](lib_path, symbol)
+                } else {
+                    return;
+                };
```
GuillaumeLagrange: Basically my above comment is: since we are resolving symbols ahead of time, is it possible that we failed to resolve the symbol earlier on but still manage to attach by name here?

GuillaumeLagrange: If that's possible, it warrants a small explanation of why it can happen

GuillaumeLagrange: else just remove the function

crates/memtrack/src/ebpf/memtrack.rs
```diff
@@ -102,50 +138,138 @@ macro_rules! attach_uprobe_uretprobe {
...
+                        "Failed to attach uretprobe to {} in {}",
+                        symbol,
+                        lib_path.display()
+                    ))?;
+                self.probes.push(link);
+
+                Ok(())
+            }
+
+            fn [<try_ $name>](
```
GuillaumeLagrange: in rust, `try_xxx` infers a `Result`. It seems we have them mixed up here

Seems like the `attach` functions should be named `try_attach_xxx`, and this function should be named something else, like `attach_if_found` or sthing

crates/memtrack/benches/attach.rs
```diff
@@ -0,0 +1,36 @@
+use memtrack::{AllocatorLib, resolve_symbol_offsets};
```
GuillaumeLagrange: shouldnt these run on vendored allocator libs as part of a `testdata` folder instead of relying on the host machine's allocator?

.github/workflows/ci.yml
```diff
@@ -134,13 +134,16 @@ jobs:
         with:
           tool: cargo-codspeed
 
+      - name: Install libbpf build deps
+        run: sudo apt-get update && sudo apt-get install -y build-essential pkgconf zlib1g-dev libbpf-dev autopoint bison flex
```
GuillaumeLagrange: there is quite a lot of duplication to install build time dependencies of memtrack across the whole workflow, cnat we create a `install-bpf-deps` local action?

Dont bundle biuld-essential with it, first check if it's required here, and if it's required here, add it as either a separate step, or an `additional-packages` action input

GuillaumeLagrange (changes requested):
small changes, olgtm

crates/memtrack/src/ebpf/memtrack.rs
not-matthias: Nope, its gone now

crates/memtrack/benches/attach.rs
not-matthias: We could, but actually since we're going to do on-demand attaching it's not really needed anymore. I'll remove it again

GuillaumeLagrange (approved):

PR #443, opened 2026-07-08
feat(memtrack): support running the memory instrument in the macro-agent sandbox

GuillaumeLagrange:
Make the memory instrument work when the runner and `codspeed-memtrack` run inside the macro-agent sandbox — unprivileged, in a user + PID namespace, with a delegated BPF token instead of root. Companion to the platform PR that sets up the token and the exe-gate.

- **Attach via `bpf()`-native links.** Switch uprobes to `uprobe_multi` and the fork tracepoint to `tp_btf`, so a BPF token can authorize them (the old `perf_event_open` path needs `CAP_PERFMON`). eBPF logic unchanged.
- **Resolve PIDs in the tracker's PID namespace.** eBPF sees global PIDs but memtrack registers namespace-local ones, so nothing matched inside the sandbox. In the init namespace this still resolves to global PIDs.
- **Accept a delegated BPF token as a privilege source.** Treat a live `LIBBPF_BPF_TOKEN_PATH` as satisfied privilege instead of requiring root / file-capabilities.

Proven end-to-end on the macro-runner: `codspeed run -m memory` on an instrumented divan bench runs inside the sandbox and uploads to staging.

Refs [COD-3047](https://linear.app/codspeed/issue/COD-3047)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,14 @@ jobs:
         # Each memtrack integration test binary runs its cases serially
         # (eBPF tracker can't overlap with itself in one process), so we
         # shard at the test-binary level to parallelize across jobs.
-        test: [c_tests, cpp_tests, rust_tests, spawn_tests]
+        test:
+          [
+            c_tests,
+            cpp_tests,
+            rust_tests,
+            spawn_tests,
+            flavor_equivalence_tests,
+          ]
     steps:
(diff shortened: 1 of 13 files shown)


GuillaumeLagrange:
The changes from this PR have been moved to #456, which combines the runner-side work for COD-3012 (hook-based CPU isolation) and COD-3047 (BPF-token memory instrument) into a single PR. Closing in favor of that one.

PR #445, opened 2026-07-08
feat(memtrack): attach allocator probes on demand

not-matthias:
Attach allocator uprobes only to libraries the tracked process tree actually maps, instead of pre-attaching every allocator discovered on the system at startup.

A BPF fentry on `security_mmap_file` signals the first mapping of each unknown executable inode. A background worker SIGSTOPs the mapping processes, classifies the file by its allocator symbols, attaches probes, and resumes them. This:

- covers `dlopen`'d allocators and statically linked allocators in spawned children (previously missed entirely)
- skips libraries the benchmark never loads
- removes the startup discovery scan (system-wide library glob + build-dir walk), which was slow on nix-heavy systems

The eager discovery path and `CODSPEED_MEMTRACK_BINARIES` are removed since the watcher supersedes them (breaking: the env var is no longer read).

The crate API is now `Tracker` (owns the attach worker) and `Session` (owns the spawned child and its event pipeline); ring buffers are polled through a generic `RingBufferPoller` with caller-supplied parsing, replacing the keepalive forwarding thread.

Review focus: the stop-the-world attach in `attach_worker.rs` (fixpoint stop + synchronous ring buffer drain) — its correctness argument is that a full `consume()` after all producers are stopped cannot miss requests.

Fixes COD-1801


crates/memtrack/src/ebpf/attach_worker.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /// Stop the worker, join it, and surface any fatal error it recorded.
+    /// Fails when the attach-request ring buffer overflowed: exec mappings were
+    /// missed, so allocator coverage would be incomplete.
+    pub(crate) fn finish(mut self) -> Result<()> {
+        self.shutdown.store(true, Ordering::SeqCst);
+        if let Some(handle) = self.handle.take() {
+            let _ = handle.join();
+        }
```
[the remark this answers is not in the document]
not-matthias: Fixed in 8bce092: join() panic is now surfaced as an error with the panic payload.

crates/memtrack/src/ebpf/c/attach.bpf.h
```diff
@@ -0,0 +1,54 @@
+/* == Exec-mapping watcher: on-demand allocator attach == */
```
GuillaumeLagrange: Slightly misleading: this is not really on demand allocator attach, this is a mmap watcher that stops the program during memmap and queues the request in a ring buffer for further processing

crates/memtrack/src/ebpf/c/memtrack.bpf.c
```diff
@@ -92,6 +92,8 @@ int tracepoint_sched_fork(struct trace_event_raw_sched_process_fork* ctx) {
     return 0;
 }
 
+#include "attach.bpf.h"
```
GuillaumeLagrange: that's slightly weird to attach.bpf.h like this no? 
is this to keep a single .bpf.c entry to easy build.rs compilation?

Maybe we could have a main.bpf.c or sthing that just includes sibling module.bpf.c files to keep things a bit tidier?

crates/memtrack/src/ebpf/c/memtrack.bpf.c
GuillaumeLagrange: completely unrelated to your code, but is this really the default behavior we want? To me, if map not initialized, the whoel thing is just not enabled

crates/memtrack/src/ebpf/events.rs
```diff
@@ -89,6 +89,31 @@ pub fn parse_event(data: &[u8]) -> Option<MemtrackEvent> {
     })
 }
 
+/// A request from the exec-mapping watcher to attach allocator probes.
+#[derive(Debug, Clone, Copy)]
+pub struct AttachRequest {
+    pub pid: u32,
+    pub dev: u64,
```
GuillaumeLagrange, line 96: we dropped the tid here, is this on purpose?

GuillaumeLagrange: If we don't need it do we need to emit it from the bpf script?

crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -1,79 +1,97 @@
...
-    /// Create a new tracker instance
+    /// Create a new tracker with on-demand allocator attach.
     ///
-    /// This will:
-    /// - Initialize the BPF subsystem
-    /// - Bump memlock limits
-    /// - Attach uprobes to all libc instances
-    /// - Attach tracepoints for fork tracking
+    /// No allocators are pre-attached: the exec-mapping watcher discovers and
+    /// attaches them as the tracked tree maps executable files.
```
GuillaumeLagrange: Overfitted comment that focuses on the difference between previous impl and this one, maybe a good candidate for the deslop skill?

crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -1,79 +1,97 @@
...
-
-        Ok(instance)
-    }
-
-    pub fn new_without_allocators() -> Result<Self> {
-        // Bump memlock limits
         Self::bump_memlock_rlimit()?;
 
         let mut bpf = MemtrackBpf::new()?;
         bpf.attach_tracepoints()?;
```
GuillaumeLagrange, line 24: would be clearer to get read of the `attach_tracepoints` indirection in favor of a direct `attach_sched_fork` call.

crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -1,79 +1,97 @@
...
-                .attach_allocator_probes(allocator.kind, &allocator.path)?;
+    /// Spawn `cmd` under tracking: the target is wrapped so it stops itself
+    /// before exec'ing, its pid is armed while stopped, then it is resumed.
+    /// The watcher observes the target's own `execve` mappings — no allocation
+    /// escapes untracked.
+    ///
+    /// `uid_gid` drops the child's privileges (a `Command`'s uid/gid cannot be
+    /// read back, so it cannot be preserved through the wrap).
+    pub fn spawn(&self, cmd: &Command, uid_gid: Option<(u32, u32)>) -> Result<Session> {
+        let mut wrapped = wrap_stopped(cmd);
```
GuillaumeLagrange, line 44: Why do we need to have the command self-stopped ??
Isnt it precisely the role of the bpf exec mapper that we've attached??

crates/memtrack/src/ebpf/spawn.rs
```diff
@@ -0,0 +1,108 @@
...
+/// happens in the already-tracked pid, so the watcher observes its mappings.
+pub fn stopped_command(program: impl AsRef<OsStr>, args: &[OsString]) -> Command {
+    let mut cmd = Command::new("sh");
+    cmd.arg("-c")
+        .arg(r#"kill -STOP "$$"; exec "$@""#)
+        .arg("sh")
+        .arg(program.as_ref())
+        .args(args);
+    cmd
+}
```
GuillaumeLagrange, line 22: I am 99% sure that we'll get both simpler code and much more reliable results by adding a ptrace(PTRACE_TRACEME) syscall, WDYT?

Did you already try it and discard it? Because while I see why we dont do a self SIGSTOP, ptrace seems more tried and tested, as well as saving us from yet another shell wrapper

GuillaumeLagrange (changes requested):
Stopping here becuase I want us to dig the TRACEME thing before leaning in the stop wrapper

crates/memtrack/src/ebpf/c/memtrack.bpf.c
not-matthias: The goal swas to separate the eBPF tracepoints into multiple files since it's getting a bit messy imo. The vision I had in mind, was to put different tracepoint categories into their own file: memory tracking, fork/attach tracking, rss tracking. etc

I used a .h since it's very easy to just include it, but I'll check if we can have multiple .c source files.

GuillaumeLagrange: I agree with the initial goal. It's also possible to include .c files btw, although not common.<br><br>But the main thing that bothers me is that we have one main file that defines a bunch of eBPF tracepoints, then includes another file, the hierarchy is a bit weird, we may want to simply create two sibling included files and keep a clean main that only includes stuff

not-matthias: Ah yes, I get what you mean. I'll see what I can do :+1:

crates/memtrack/src/ebpf/tracker.rs
not-matthias: Well, we're only doing the fork/exec tracking for `tracked_pids` which we'll have to setup at some point. So there's a small race condition window at the start that drops allocations. 

But maybe we can use `pre_exec` for this

crates/memtrack/src/ebpf/c/memtrack.bpf.c
not-matthias: Good point, changed it

crates/memtrack/src/ebpf/spawn.rs
not-matthias: The main reason we need this is to not have to manually parse the /proc/maps mappings to find the static/dynamic allocators after the execve if we use ptrace. 

So basically what we're doing right here is: 
1. Spawn `sh`, stop it
2. Track this pid in eBPF
3. Exec the actual target program -> eBPF can now use the same logic to detect new mappings/forks/execs

 If we wanted to use ptrace, we'd still do `sh -c 'exec "$@"'` and then use PTRACE_TRACEME to stop it in `pre_exec`. I don't like this as much since it separates the logic and makes it harder to understand for (almost) no benefit. 

Wdyt?

not-matthias: As discussed, due to `pre-exec` not working on statically linked binaries we'll continue to use the current logic.

crates/memtrack/src/ebpf/c/memtrack.bpf.c
not-matthias: Created a separate PR for this: https://github.com/CodSpeedHQ/codspeed/pull/455 (i'll rebase it once it's been merged)

GuillaumeLagrange (approved):
lgtm, I recommend you rebase on top of CodSpeedHQ/codspeed#455 before merging though 😅

not-matthias:
Yeah, that was the plan :)

PR #453, opened 2026-07-13
feat(memtrack): collect RSS via rss_stat and folio-rmap reconstruction

not-matthias:
## Summary

Adds RSS (resident set size) collection to `memtrack`, in two layers:

1. **Authoritative RSS** from the kernel's `kmem:rss_stat` tracepoint — absolute per-mm resident bytes (anon/file/shmem/swap), latest-wins.
2. **Reconstructed anonymous RSS** from raw kernel folio-rmap events: `fentry` hooks on the anon folio-rmap add/remove functions emit signed page-count deltas, so anon RSS can be rebuilt over time as `Σ(add − remove) × PAGE_SIZE`.

The reconstruction is a **total anon RSS delta-sum**, not a per-vaddr resident map — the kernel remove hook (`folio_remove_rmap_ptes`) carries no address, so removals can't be attributed to a vaddr (the add hooks' faulting vaddr is emitted for observability only).

## Commits

- `feat(memtrack): track RSS via kmem:rss_stat tracepoint` — the baseline: `EVENT_TYPE_RSS` contract, `MemtrackEventKind::Rss`, parser arm, writer bench case, gated integration test.
- `feat(memtrack): reconstruct anon RSS from gated folio rmap fentry hooks` — `EVENT_TYPE_RMAP_ANON` + `RmapAnon` event, five `fentry` programs (add_new / add_ptes / remove_ptes / remove_pmd / remove_pud), CO-RE folio helpers, and the load/attach gating.
- `test(memtrack): validate anon RSS reconstruction against rss_stat` — ramps anon RSS via `mmap`/`munmap` and asserts the reconstructed estimate tracks the `rss_stat` MM_ANONPAGES peak within 25%.

## What's on by default vs gated

- **`rss_stat` tracepoint: always on.** Emitting `Rss` events is the intended new default behavior introduced by this change — the RSS tracepoint is not gated.
- **`RmapAnon` folio-rmap `fentry` programs: off by default**, gated behind `CODSPEED_MEMTRACK_TRACK_RMAP=1`. When the flag is unset they are `set_autoload(false)` before load and never attached, so:
  - the skeleton still loads on any kernel (a missing `fentry` BTF target would otherwise fail the whole load), and
  - the folio-rmap reconstruction path stays out of production (`--mode memory`) and out of the existing test suites — no `RmapAnon` events are produced by default.

## Verification

Run in a privileged, `--pid=host` container sharing the host kernel (7.0.12):

- **Reconstruction (flag on):** ✅ `real anon amplitude = 64 MiB, estimated peak = 64 MiB` (ratio 1.00).
- **Regression (`rss_tests`, flag unset):** ✅ passes — no `RmapAnon` events, folio-rmap programs stay unloaded.
- Parser unit tests, `cargo fmt`, and `clippy` clean.
- Kernel BTF signatures for all five `folio_*_rmap*` functions verified to match the `BPF_PROG` arg layouts.

## Review notes (draft)

- **`track_command` ordering:** the shared test helper spawns the child before `enable()`/`track(root_pid)`. In practice the child's `fork→execve→ld.so→libc-init` far outlasts the two BPF-map updates, so tracking is armed before the workload allocates (both fixtures captured full event streams). Flagging in case we'd prefer a leading settle-`usleep` in the fixtures or an enable-before-spawn change in the helper.
- **`PAGE_SIZE`:** hardcoded to 4096 (correct on x86_64). On a 16K/64K-page arm64 runner the estimate would need `sysconf(_SC_PAGESIZE)`; `rss_stat` is already in bytes and unaffected. Happy to switch to `sysconf` if these tests run on arm64 CI.
- The chart/inspection tooling used during development lives outside the tree (dev artifact) and is not part of this PR.

### DIFF
--- a/crates/memtrack/src/ebpf/c/rss.bpf.h
+++ b/crates/memtrack/src/ebpf/c/rss.bpf.h
@@ -5,6 +5,74 @@
 #include "utils/event_helpers.h"
 #include "utils/process_tracking.h"
 
+/* (rss_stat mm_id << 32 | member) -> {owning tgid, last in-context size}. Keyed per
+ * counter so an external (curr==0) update is attributed only once that mm/member was
+ * established in-context. An external event may only lower a counter: any size above
+ * the last in-context value is dropped, so neither a stale/racing reclaim read nor an
+ * mm_id hash collision with another task can invent a peak. LRU eviction + re-seeding
+ * on the owner's next in-context event covers hash reuse, so no teardown hook needed. */
+struct rss_owner {
+    __u32 pid;
+    __u64 size;
+};
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __uint(max_entries, 40960);
+    __type(key, __u64);
+    __type(value, struct rss_owner);
+} rss_counter_owner SEC(".maps");
+
+static __always_inline int submit_rss_event(__u32 owner_pid, __s32 member, __u64 size) {
+    SUBMIT_EVENT_AS(owner_pid, EVENT_TYPE_RSS, {
+        e->data.rss.member = member;
+        e->data.rss.size = size;
+    });
+}
+
+SEC("tracepoint/kmem/rss_stat")
+int tracepoint_rss_stat(struct trace_event_raw_rss_stat* ctx) {
+    /* Swap is out of scope: MM_SWAPENTS counts entries that are no longer
+     * resident, and nothing downstream reports them. */
+    if (ctx->member == MM_SWAPENTS) {
+        return 0;
+    }
+
+    __u32 cur = current_tgid();
+    __u64 key = ((__u64)ctx->mm_id << 32) | (__u32)ctx->member;
+    __u64 size = ctx->size;
+    __u32 owner;
+
+    if (ctx->curr) {
+        if (!is_tracked(cur)) {
+            return 0;
+        }
+        owner = cur;
+        struct rss_owner state = {.pid = cur, .size = size};
+        bpf_map_update_elem(&rss_counter_owner, &key, &state, BPF_ANY);
+    } else {
+        struct rss_owner* found = bpf_map_lookup_elem(&rss_counter_owner, &key);
+        if (!found) {
+            return 0;
+        }
+        owner = found->pid;
+        /* The owner's own teardown also presents as curr==0 (current->mm is cleared
+         * on exit), so drop it. Genuine external actors (reclaim, another process's
+         * madvise) run in a different task, so cur != owner. */
+        if (cur == owner) {
+            return 0;
+        }
+        /* An external actor may only lower a counter. A larger value is a stale
+         * reclaim read or an mm_id hash collision with another task; dropping it
+         * keeps the reconstructed peak identical to the in-context timeline. */
+        if (size > found->size) {
+            return 0;
+        }
+        found->size = size;
+    }
+
+    return submit_rss_event(owner, ctx->member, size);
+}
+
 /* == Process lifecycle events ==
  *
  * FORK lets userland seed a child's RSS from its parent at fork time: the
--- a/crates/memtrack/src/ebpf/c/rss.bpf.h
+++ b/crates/memtrack/src/ebpf/c/rss.bpf.h
@@ -3,16 +3,19 @@
 
 #include "event.h"
 #include "utils/event_helpers.h"
+#include "utils/mm_ownership.h"
 #include "utils/process_tracking.h"
 
-/* (rss_stat mm_id << 32 | member) -> {owning tgid, last in-context size}. Keyed per
- * counter so an external (curr==0) update is attributed only once that mm/member was
- * established in-context. An external event may only lower a counter: any size above
- * the last in-context value is dropped, so neither a stale/racing reclaim read nor an
- * mm_id hash collision with another task can invent a peak. LRU eviction + re-seeding
- * on the owner's next in-context event covers hash reuse, so no teardown hook needed. */
+/* Key: rss_stat's mm_id in the upper 32 bits, the counter member in the lower.
+ * Value: the tgid that last reported this counter from its own context, the mm it
+ * held then, and the last size accepted for it.
+ *
+ * mm_id is a hash, so an out-of-context report (curr == 0) is accepted only while
+ * mm_by_pid still binds that tgid to the seeded mm, and only when it lowers the
+ * counter: a hash collision or a stale reclaim read must never invent a peak. */
 struct rss_owner {
     __u32 pid;
+    __u64 mm;
     __u64 size;
 };
 struct {
@@ -20,7 +23,7 @@ struct {
     __uint(max_entries, 40960);
     __type(key, __u64);
     __type(value, struct rss_owner);
-} rss_counter_owner SEC(".maps");
+} rss_owner_by_counter SEC(".maps");
 
 static __always_inline int submit_rss_event(__u32 owner_pid, __s32 member, __u64 size) {
     SUBMIT_EVENT_AS(owner_pid, EVENT_TYPE_RSS, {
@@ -47,10 +50,15 @@ int tracepoint_rss_stat(struct trace_event_raw_rss_stat* ctx) {
             return 0;
         }
         owner = cur;
-        struct rss_owner state = {.pid = cur, .size = size};
-        bpf_map_update_elem(&rss_counter_owner, &key, &state, BPF_ANY);
+        /* curr == 1 means current->mm is the counter's mm. mm_by_pid is maintained
+         * here too because the rmap hooks may not be attached. */
+        struct task_struct* task = bpf_get_current_task_btf();
+        __u64 mm = (__u64)BPF_CORE_READ(task, mm);
+        struct rss_owner state = {.pid = cur, .mm = mm, .size = size};
+        bpf_map_update_elem(&rss_owner_by_counter, &key, &state, BPF_ANY);
+        rebind_pid_mm(cur, mm);
     } else {
-        struct rss_owner* found = bpf_map_lookup_elem(&rss_counter_owner, &key);
+        struct rss_owner* found = bpf_map_lookup_elem(&rss_owner_by_counter, &key);
         if (!found) {
             return 0;
         }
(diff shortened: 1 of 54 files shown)


not-matthias:
@greptileai pause reviews for now until i enable them again

not-matthias:
@greptileai review again, focus on the eBPF code (mostly RMAP tracking), be exhaustive

crates/memtrack/src/ebpf/c/rss.bpf.h
```diff
@@ -0,0 +1,401 @@
...
+        e->data.rss.member = member;
+        e->data.rss.size = size;
+    });
+}
+
+SEC("tracepoint/kmem/rss_stat")
+int tracepoint_rss_stat(struct trace_event_raw_rss_stat* ctx) {
+    if (ctx->member == MM_SWAPENTS) {
+        return 0;
+    }
```
GuillaumeLagrange, line 41: Explicitly state we do not support swap

crates/memtrack/src/ebpf/c/rss.bpf.h
```diff
@@ -0,0 +1,401 @@
...
+#define __RSS_BPF_H__
+
+#include "event.h"
+#include "utils/event_helpers.h"
+#include "utils/process_tracking.h"
+
+/* (rss_stat mm_id << 32 | member) -> {owning tgid, its mm when seeded, last
+ * in-context size}. mm_id is a hash, so an external (curr==0) update is accepted
+ * only while pid_mm still binds the owner to the seeded mm, and only to lower
+ * the counter — collisions and stale reads must not invent a peak. */
```
GuillaumeLagrange: I have a hard time understanding this comment, can we rephrase?

crates/memtrack/src/ebpf/c/rss.bpf.h
```diff
@@ -0,0 +1,401 @@
...
+ *
+ * pid_mm must not use LRU eviction: losing the inverse binding would leave exec
+ * and exit unable to remove the forward entry. */
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __uint(max_entries, 10240);
+    __type(key, __u64);
+    __type(value, __u32);
+} mm_owner SEC(".maps");
+BPF_HASH_MAP(pid_mm, __u32, __u64, 10240);
```
GuillaumeLagrange: Can we rename it to better reflect what's the data and what's the key? A convention I like is `{object}_by_{key}`

For both this and the rss_counter_owner it could be good!

crates/memtrack/src/ebpf/c/rss.bpf.h
```diff
@@ -0,0 +1,401 @@
...
+        return BPF_CORE_READ(folio, flags).f;
+    }
+    return BPF_CORE_READ((struct folio___legacy*)folio, flags);
+}
+
+/* Kernels < 6.6 store the large-folio order in a dedicated byte instead of
+ * the low byte of _flags_1. */
+struct folio___order_byte {
+    unsigned char _folio_order;
+} __attribute__((preserve_access_index));
```
GuillaumeLagrange: this could benefit form being in a dedicated legacy or kernel support file

crates/memtrack/src/ebpf/c/rss.bpf.h
```diff
@@ -0,0 +1,401 @@
...
+     * the tracker's namespace, which the tracepoint's raw pid field cannot
+     * give. */
+    __u32 child_pid = task_ns_tgid(child);
+    if (!child_pid) {
+        return 0;
+    }
+    track_child(child_pid, parent_pid);
+
+    SUBMIT_EVENT_AS(child_pid, EVENT_TYPE_FORK, { e->data.fork.parent_pid = parent_pid; });
+}
```
GuillaumeLagrange: Does not belong to the rss.bpf, it belongs to a specific fork/exec events watcher module

crates/memtrack/src/ebpf/c/utils/process_tracking.h
```diff
@@ -44,27 +44,4 @@ static __always_inline void track_child(__u32 child_pid, __u32 parent_pid) {
...
- * be), and its arguments are task_struct pointers rather than a flattened trace
- * event, so PIDs resolve through the tracker's namespace.
- *
- * https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_TRACING/
- */
-SEC("tp_btf/sched_process_fork")
-int BPF_PROG(tracepoint_sched_fork, struct task_struct* parent, struct task_struct* child) {
-    follow_fork(task_ns_tgid(parent), task_ns_tgid(child));
-    return 0;
-}
```
GuillaumeLagrange, line 68: Related to the other comments, why was this removing from the process_tracking code and moved into the rss one?

crates/memtrack/src/ebpf/memtrack/maps.rs
```diff
@@ -67,6 +67,50 @@ impl MemtrackBpf {
...
+                .collect(),
+        })
+    }
+}
+
+/// Live ownership bindings: `mm_owner` (`mm_struct` pointer -> owning pid) and its
+/// inverse `pid_mm`, which foreign-actor rmap attribution validates against.
+pub struct OwnershipMaps {
+    pub mm_owner: Vec<(u64, u32)>,
+    pub pid_mm: Vec<(u32, u64)>,
```
GuillaumeLagrange: Same remark for map names here, `{object}_by_{key}` is so much clearer IMO

crates/memtrack/src/ebpf/c/utils/event_helpers.h
```diff
@@ -40,14 +40,19 @@ static __always_inline __u64* take_param(void* map) {
...
 
-#define SUBMIT_EVENT(evt_type, fill_data)                               \
+/* Submission is split into two classes:
+ *  - lifetime events (rss_stat, rmap, fork/exec/exit): emitted whenever the
+ *    pid is tracked, ignoring the enable toggle. The parser reconstructs
+ *    absolute per-process state from these, so it needs every event from
+ *    process birth — a delta stream that starts mid-life can never recover
+ *    the resident baseline faulted before enable.
+ *  - allocator events (malloc/free/mmap/...): high-volume and only meaningful
+ *    inside a measurement window, so they stay behind is_enabled().
```
GuillaumeLagrange: `gated events`, just have this doc mention that some events are gated by a toggling flag, with the allocator ones being the only ones for now

crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -12,40 +12,62 @@ use std::sync::mpsc;
...
+    allocators: bool,
 }
 
 impl Tracker {
     /// Create a new tracker. The exec-mapping watcher discovers and attaches
     /// allocator probes as the tracked process tree maps executable files.
     pub fn new() -> Result<Self> {
-        Self::with_bpf(MemtrackBpf::new()?)
+        let track_rmap = Self::track_rmap_from_env();
+        Self::build(MemtrackBpf::new_with_rmap(track_rmap)?, true)
```
GuillaumeLagrange, line 23: I think we should have a gate here on the kernel version, else we will have warnings when running memtrack on older kernels because we'll try and fail hooking into the rmap kernel functions.<br><br>Additionally, do you think we could centralize this kind of kernel version/feature oin some way? I fear this will quickly become unmaintainable with the way we do it for `core` and `pud` thing.

Let's not overengineer something, but I think we could have some sort of setting where

* a minimal kernel version is defined to attempt to hook
* we should handle failure to hook gracefully with warnings
* this information is somehow centralized instead of scatered accross rust struct's docstrings and/or bpf code comment

GuillaumeLagrange (approved):
olgtm, obviously quite a big chunk to chew, but looks mostly fine to me. Main point of concern is the maintainability of "which function can be track for which kernel version" which IMO could be improved.

crates/memtrack/src/ebpf/c/rss.bpf.h
GuillaumeLagrange: Same for the sched_process_exec probe, WDYT?

crates/memtrack/src/ebpf/tracker.rs
not-matthias: Added `RmapSupport { Unsupported, Core, CoreAndPud }`. We can extend it or add similar features later on.

crates/memtrack/src/ebpf/c/rss.bpf.h
not-matthias: Created a new process_tracking.bpf.h to prevent cyclic inclusions.

crates/memtrack/src/ebpf/c/utils/process_tracking.h
not-matthias: Good question, I think mainly for simplicity + getting rid of the cyclic inclusion of the process_tracking.h. Fixed it now

crates/memtrack/src/ebpf/memtrack/tracking.rs
```diff
@@ -3,10 +3,39 @@ use crate::prelude::*;
...
+    attach_tracepoint!(rss_stat);
+    attach_tracepoint!(task_newtask);
+    attach_tracepoint!(sched_process_exec);
+    attach_tracepoint!(sched_process_exit);
 
     pub fn attach_tracepoints(&mut self) -> Result<()> {
-        self.attach_sched_fork()?;
+        self.attach_task_newtask()?;
+        self.attach_sched_process_exec()?;
+        self.attach_sched_process_exit()?;
```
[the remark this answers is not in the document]
not-matthias: If we can't attach to those, we can't really run because we need the fork/clone/exec information for other cases. Disregarding this comment

not-matthias:
@greptileai review

crates/memtrack/src/ebpf/memtrack/rmap.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    fn for_version(version: KernelVersion) -> Self {
+        if version < KernelVersion::new(6, 8) {
+            return Self::Unsupported;
+        }
+        if version < KernelVersion::new(6, 15) {
+            return Self::Core;
+        }
+        Self::CoreAndPud
```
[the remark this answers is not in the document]
not-matthias: CoreAndPud must exist on 6.15+

PR #456, opened 2026-07-15
feat: run CPU isolation and the memory instrument inside the macro-agent sandbox

GuillaumeLagrange:
Runner-side changes to make CPU isolation and the memory instrument work
unprivileged inside the macro-agent sandbox. Two issues, combined here because
they share the sandbox's privilege model and the same branch.

**CPU isolation (COD-3012).** Drop the runner's built-in `systemd-run` /
`CGROUP:` isolation for a hook-based one: an `Isolation` type invokes
`codspeed-{pre,wrap,post}-bench` and stays oblivious to how cores are attributed.
If `codspeed-wrap-bench` exists it takes the (unprivileged) hook path; otherwise
it falls back to `systemd-run --scope --slice=codspeed.slice`, so non-sandbox
hosts are unchanged.

**Memory instrument (COD-3047).** Let memtrack run off a delegated BPF token
instead of root: attach via `bpf()`-native links (`uprobe_multi` + `tp_btf`),
resolve tracked PIDs in the tracker's PID namespace, and accept
`LIBBPF_BPF_TOKEN_PATH` as a privilege source.

Closes COD-3012
Closes COD-3047

### DIFF
--- a/crates/memtrack/src/ebpf/attach_worker.rs
+++ b/crates/memtrack/src/ebpf/attach_worker.rs
@@ -232,8 +232,38 @@ impl Worker {
             ),
         };
 
-        let Ok(lib) = AllocatorLib::from_path_static(&mapping.attach_path) else {
-            debug!("on-demand: not an allocator: {}", mapping.display);
+        let candidates = mapping.candidate_paths();
+        let mut lib = None;
+        let mut unreadable = Vec::new();
+        for path in &candidates {
+            match AllocatorLib::from_path_static(path) {
+                Ok(found) => {
+                    lib = Some(found);
+                    break;
+                }
+                // Classification reads the whole file, so a failed open is
+                // indistinguishable from a readable non-allocator. Probe
+                // openability to tell them apart.
+                Err(_) => {
+                    if let Err(e) = std::fs::File::open(path) {
+                        unreadable.push(format!("{}: {e}", path.display()));
+                    }
+                }
+            }
+        }
+
+        let Some(lib) = lib else {
+            // A mapping no path can reach costs coverage of that one library,
+            // which is not worth killing the run over.
+            if unreadable.len() == candidates.len() {
+                warn!(
+                    "no readable path for mapping {} in pid {}, allocator coverage may be incomplete",
+                    mapping.display, req.pid
+                );
+                debug!("unreadable paths: {}", unreadable.join(", "));
+            } else {
+                debug!("on-demand: not an allocator: {}", mapping.display);
+            }
             return Ok(true);
         };
 
--- a/crates/memtrack/src/ebpf/c/attach.h
+++ b/crates/memtrack/src/ebpf/c/attach.h
@@ -31,8 +31,7 @@ int BPF_PROG(watch_exec_mmap, struct file* file, unsigned long prot, unsigned lo
         return 0;
     }
 
-    __u64 pid_tgid = bpf_get_current_pid_tgid();
-    __u32 tgid = pid_tgid >> 32;
+    __u32 tgid = current_tgid();
     if (!is_tracked(tgid)) {
         return 0;
     }
--- a/crates/memtrack/src/ebpf/c/memtrack_token.bpf.c
+++ b/crates/memtrack/src/ebpf/c/memtrack_token.bpf.c
@@ -1,0 +1,7 @@
+/* Token variant: uprobes attach as uprobe_multi links, so every attach goes
+ * through bpf() and a delegated BPF token can authorize it without host
+ * privileges. Requires kernel >= 6.6. See utils/variant.h for why classic
+ * uprobes cannot be delegated. Program bodies live in main.bpf.c; only the
+ * SEC() annotations differ, keyed on this define. */
+#define MEMTRACK_BPF_VARIANT_TOKEN 1
+#include "main.bpf.c"
--- a/crates/memtrack/src/ebpf/c/utils/event_helpers.h
+++ b/crates/memtrack/src/ebpf/c/utils/event_helpers.h
@@ -21,16 +21,18 @@ static __always_inline long wake_flags(void) {
 }
 
 static __always_inline int store_param(void* map, __u64 value) {
-    __u64 tid = bpf_get_current_pid_tgid();
-    __u32 pid = tid >> 32;
-    if (is_tracked(pid)) {
+    /* Key by the tid: unique per thread, so it survives the entry/exit pair even
+     * when several threads are inside the same allocator call. */
+    struct task_ids ids = current_task_ids();
+    __u64 tid = ids.tid;
+    if (is_tracked(ids.tgid)) {
         bpf_map_update_elem(map, &tid, &value, BPF_ANY);
     }
     return 0;
 }
 
 static __always_inline __u64* take_param(void* map) {
-    __u64 tid = bpf_get_current_pid_tgid();
+    __u64 tid = current_tid();
     __u64* value = bpf_map_lookup_elem(map, &tid);
     if (value) {
         bpf_map_delete_elem(map, &tid);
@@ -40,10 +42,9 @@ static __always_inline __u64* take_param(void* map) {
 
 #define SUBMIT_EVENT(evt_type, fill_data)                               \
     {                                                                   \
-        __u64 tid = bpf_get_current_pid_tgid();                         \
-        __u32 pid = tid >> 32;                                          \
+        struct task_ids ids = current_task_ids();                       \
                                                                         \
-        if (!is_tracked(pid) || !is_enabled()) {                        \
+        if (!is_tracked(ids.tgid) || !is_enabled()) {                   \
             return 0;                                                   \
         }                                                               \
                                                                         \
@@ -58,8 +59,8 @@ static __always_inline __u64* take_param(void* map) {
         }                                                               \
                                                                         \
         e->header.timestamp = bpf_ktime_get_ns();                       \
-        e->header.pid = pid;                                            \
-        e->header.tid = tid & 0xFFFFFFFF;                               \
+        e->header.pid = ids.tgid;                                       \
+        e->header.tid = ids.tid;                                        \
         e->header.event_type = evt_type;                                \
                                                                         \
         fill_data;                                                      \
--- a/crates/memtrack/src/ebpf/c/utils/process_tracking.h
+++ b/crates/memtrack/src/ebpf/c/utils/process_tracking.h
@@ -2,6 +2,7 @@
 #define __PROCESS_TRACKING_H__
 
 #include "map_helpers.h"
+#include "variant.h"
 
 BPF_HASH_MAP(tracked_pids, __u32, __u8, 10000);
 BPF_HASH_MAP(pids_ppid, __u32, __u32, 10000);
@@ -43,15 +44,26 @@ static __always_inline void track_child(__u32 child_pid, __u32 parent_pid) {
     bpf_map_update_elem(&pids_ppid, &child_pid, &parent_pid, BPF_ANY);
 }
 
-SEC("tracepoint/sched/sched_process_fork")
-int tracepoint_sched_fork(struct trace_event_raw_sched_process_fork* ctx) {
-    __u32 parent_pid = ctx->parent_pid;
-    __u32 child_pid = ctx->child_pid;
-
+/* Record a parent→child fork so the child inherits the parent's tracked state. */
+static __always_inline void follow_fork(__u32 parent_pid, __u32 child_pid) {
+    if (parent_pid == 0 || child_pid == 0) {
+        return;
+    }
     if (is_tracked(parent_pid)) {
         track_child(child_pid, parent_pid);
     }
+}
 
+/* tp_btf rather than a classic tracepoint on two counts: it attaches with
+ * BPF_RAW_TRACEPOINT_OPEN, so a token can delegate it (perf_event_open() cannot
+ * be), and its arguments are task_struct pointers rather than a flattened trace
+ * event, so PIDs resolve through the tracker's namespace.
+ *
+ * https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_TRACING/
+ */
+SEC("tp_btf/sched_process_fork")
+int BPF_PROG(tracepoint_sched_fork, struct task_struct* parent, struct task_struct* child) {
+    follow_fork(task_ns_tgid(parent), task_ns_tgid(child));
     return 0;
 }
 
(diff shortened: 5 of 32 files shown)


GuillaumeLagrange:
@greptileai review

crates/memtrack/src/ebpf/memtrack/mod.rs
```diff
@@ -8,17 +8,73 @@ use std::path::Path;
...
+/// its `bpf()` calls; its presence is also what tells us to use the bpf()-link
+/// attach paths instead of the perf-based ones.
+fn has_delegated_bpf_token() -> bool {
+    std::env::var_os("LIBBPF_BPF_TOKEN_PATH")
+        .is_some_and(|p| !p.is_empty() && std::path::Path::new(&p).is_dir())
+}
+
+/// Which set of attach mechanisms a loaded skeleton uses.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Flavor {
```
not-matthias: nit: When I read flavor I think of 🧁

Wdyt of MemtrackVariant or BpfVariant?

crates/memtrack/src/ebpf/memtrack/mod.rs
```diff
@@ -74,23 +130,64 @@ impl ResolvedSymbols {
...
-    probes: Vec<Link>,
+    pub(super) skel: Skel,
+    pub(super) probes: Vec<Link>,
 }
 
 impl MemtrackBpf {
+    /// Load the skeleton, selecting the attach flavor from the environment: a
+    /// delegated BPF token means we are the sandboxed workload and must use the
+    /// token path; otherwise the perf path, which also supports kernels
+    /// predating uprobe_multi.
```
not-matthias: this comment can be simplified/deslopped

crates/memtrack/src/ebpf/memtrack/mod.rs
```diff
@@ -74,23 +130,64 @@ impl ResolvedSymbols {
...
+        } else {
+            Flavor::Perf
+        };
+        Self::with_flavor(flavor)
+    }
+
+    /// Load the skeleton for a specific attach flavor, bypassing environment
+    /// detection. Used by tests to exercise either path directly; both attach
+    /// fine on a privileged kernel without a token (the token only authorizes
+    /// `bpf()` from inside the unprivileged sandbox).
```
not-matthias: this comment is also a bit too verbose imo

> ```
> the token only authorizes
>     /// `bpf()` from inside the unprivileged sandbox
> ```

would it make sense to also test this behavior here to ensure it works? wdyt?

crates/memtrack/src/ebpf/memtrack/mod.rs
```diff
@@ -8,17 +8,73 @@ use std::path::Path;
...
+/// Set the PID-namespace rodata on an open skeleton, so both flavors resolve
+/// PIDs the same way. `$open` is the open skeleton, of either flavor.
+macro_rules! set_pidns_rodata {
+    ($open:expr, $ids:expr) => {
+        if let (Some((dev, ino)), Some(rodata)) = ($ids, $open.maps.rodata_data.as_deref_mut()) {
+            rodata.target_pidns_dev = dev;
+            rodata.target_pidns_ino = ino;
+        }
+    };
+}
```
not-matthias: nit: since this macro is only used inside the with_flavor func, can we just move it there to not have it in the global scope (even if it's not exported)

crates/memtrack/src/ebpf/c/utils/flavor.h
```diff
@@ -0,0 +1,61 @@
...
+ * hook as tp_btf, both authorized through bpf(). Without it they use the classic
+ * perf-based uprobe/tracepoint paths, which work on kernels predating
+ * uprobe_multi (< 6.6) but cannot be delegated into the sandbox. */
+#ifdef MEMTRACK_BPF_LINKS
+#define UPROBE_SEC "uprobe.multi"
+#define URETPROBE_SEC "uretprobe.multi"
+#else
+#define UPROBE_SEC "uprobe"
+#define URETPROBE_SEC "uretprobe"
+#endif
```
not-matthias: As discussed, let's check if we need uprobe.multi

crates/memtrack/src/ebpf/c/memtrack_token.bpf.c
```diff
@@ -0,0 +1,6 @@
+/* BPF-token attach variant: uprobe_multi links + tp_btf fork hook, authorized
+ * through bpf() so a delegated token can load them inside the sandbox. Requires
+ * a kernel with uprobe_multi (>= 6.6). The program bodies live in the shared
+ * main.bpf.c; only the SEC() annotations differ, keyed on this define. */
+#define MEMTRACK_BPF_LINKS 1
```
not-matthias: i'm not sure I get the naming of `MEMTRACK_BPF_LINKS`, wdyt of `MEMTRACK_FLAVOR_IS_TOKEN` or `MEMTRACK_FLAVOR?`

crates/memtrack/src/ebpf/c/memtrack_perf.bpf.c
```diff
@@ -0,0 +1,5 @@
+/* Classic perf attach variant: perf-based uprobe/uretprobe + a perf
+ * sched_process_fork tracepoint. Works on kernels predating uprobe_multi but
+ * needs CAP_PERFMON in the init user namespace, so it cannot be delegated into
+ * the sandbox. The program bodies live in the shared main.bpf.c. */
```
not-matthias: > ```
>  so it cannot be delegated into the sandbox
> ```

nit: we have a lot of references to "the sandbox" here in the code and comments, i'm not sure if we should link it as closely. we could keep it more abstract and call it "inside containers"

crates/memtrack/src/ebpf/c/memtrack_perf.bpf.c
```diff
@@ -0,0 +1,5 @@
+/* Classic perf attach variant: perf-based uprobe/uretprobe + a perf
```
not-matthias: Does "perf" make sense here for the name of the variant/file? I associate it more with the perf tool.

crates/memtrack/tests/flavor_equivalence_tests.rs
```diff
@@ -0,0 +1,123 @@
+//! Verifies the two eBPF attach flavors produce equivalent tracking results.
```
not-matthias: Wouldn't it make more sense to simply run the existing tests both with flavor::token and flavor::perf? because then we can reuse the existing test fixtures rather than causing a split and having to test everything separately.

it *should* result in the same snapshots, right?

crates/memtrack/src/ebpf/c/utils/process_tracking.h
```diff
@@ -43,16 +44,34 @@ static __always_inline void track_child(__u32 child_pid, __u32 parent_pid) {
...
+static __always_inline void follow_fork(__u32 parent_pid, __u32 child_pid) {
+    if (parent_pid == 0 || child_pid == 0) {
+        return;
+    }
     if (is_tracked(parent_pid)) {
         track_child(child_pid, parent_pid);
     }
+}
 
+#ifdef MEMTRACK_BPF_LINKS
```
not-matthias: no idea if possible, but could we use the tp_btf solution for both?

crates/memtrack/src/ebpf/c/attach.h
```diff
@@ -31,8 +31,7 @@ int BPF_PROG(watch_exec_mmap, struct file* file, unsigned long prot, unsigned lo
         return 0;
     }
 
-    __u64 pid_tgid = bpf_get_current_pid_tgid();
-    __u32 tgid = pid_tgid >> 32;
+    __u32 tgid = current_pid();
```
not-matthias: tgid is actually the wrong name here, this should be named `pid`

crates/memtrack/src/ebpf/c/utils/flavor.h
```diff
@@ -0,0 +1,61 @@
...
+
+/* PID namespace the userspace tracker resolves PIDs in. When set (ino != 0),
+ * PIDs are read relative to this namespace so tracking works when the tracker
+ * runs inside a PID namespace (e.g. the macro-agent sandbox) while eBPF sees
+ * global PIDs. Zero means "use global PIDs" — the default outside a sandbox. */
+const volatile __u64 target_pidns_dev = 0;
+const volatile __u64 target_pidns_ino = 0;
+
+/* Current task's PID in the configured namespace (or global when unset). */
+static __always_inline __u32 current_pid(void) {
```
not-matthias: don't we need to return both the tid and pid for the target namespace?

we could mimic the style of `bpf_get_current_pid_tgid` and have `memtrack_current_pid_tgid` or something like that

not-matthias (changes requested):
Overall, already pretty good. I think we can improve the naming a bit and let's check if we can keep the tests unified.

Lmk if you have some other thoughts on some comments.

crates/memtrack/src/ebpf/memtrack/mod.rs
GuillaumeLagrange: Changed to `BpfVariant`

crates/memtrack/src/ebpf/memtrack/mod.rs
GuillaumeLagrange: Removed these comments and made the explanation more direct, I've also added more context to the various tickets for the design decision

crates/memtrack/src/ebpf/c/utils/flavor.h
GuillaumeLagrange: We do need it, extended explanation can be found on the `cod-3047` ticket

crates/memtrack/src/ebpf/c/memtrack_token.bpf.c
GuillaumeLagrange: changed it to `MEMTRACK_BPF_VARIANT_TOKEN`

crates/memtrack/src/ebpf/c/memtrack_perf.bpf.c
GuillaumeLagrange: removed all references to "the sandbox" you're right

crates/memtrack/src/ebpf/c/memtrack_perf.bpf.c
GuillaumeLagrange: Changed it to `Legacy`, lmk if you'd prefer something else

crates/memtrack/tests/flavor_equivalence_tests.rs
GuillaumeLagrange: Done, all tests now do this, removed the dedicated tests

crates/memtrack/src/ebpf/c/utils/process_tracking.h
GuillaumeLagrange: changed it for both, it's been supported since kernel 5.5 so we should be good

crates/memtrack/src/ebpf/c/utils/event_helpers.h
```diff
@@ -40,8 +41,8 @@ static __always_inline __u64* take_param(void* map) {
 
 #define SUBMIT_EVENT(evt_type, fill_data)                               \
     {                                                                   \
-        __u64 tid = bpf_get_current_pid_tgid();                         \
-        __u32 pid = tid >> 32;                                          \
+        __u32 tid = current_tid();                                      \
+        __u32 pid = current_tgid();                                     \
```
not-matthias: This is called on each allocator call, and since it's already quite slow we should try to do as little work as possible. We should only ever call `bpf_get_ns_current_pid_tgid` once and then extract the pid/tgid from it. 

I'm not sure what the performance impact of this would be (maybe it's negligible). Maybe we should also consider setting up some benches running eBPF (we can do this later)

crates/memtrack/src/ebpf/memtrack/maps.rs
```diff
@@ -81,13 +72,12 @@ impl MemtrackBpf {
...
-            .skel
+        let value = with_skel!(self, skel => skel
             .maps
             .dropped_events
-            .lookup(&key.to_le_bytes(), libbpf_rs::MapFlags::ANY)
-            .context("Failed to read dropped_events counter")?
-            .ok_or_else(|| anyhow!("dropped_events slot 0 missing"))?;
+            .lookup(&key.to_le_bytes(), libbpf_rs::MapFlags::ANY))
+        .context("Failed to read dropped_events counter")?
+        .ok_or_else(|| anyhow!("dropped_events slot 0 missing"))?;
```
not-matthias: nit: formatting looks a bit off here

crates/memtrack/src/ebpf/memtrack/maps.rs
```diff
@@ -53,24 +44,24 @@ impl MemtrackBpf {
...
-            .skel
+        let value = with_skel!(self, skel => skel
             .maps
             .attach_request_dropped
-            .lookup(&key.to_le_bytes(), libbpf_rs::MapFlags::ANY)
-            .context("Failed to read attach_request_dropped counter")?
-            .ok_or_else(|| anyhow!("attach_request_dropped slot 0 missing"))?;
+            .lookup(&key.to_le_bytes(), libbpf_rs::MapFlags::ANY))
+        .context("Failed to read attach_request_dropped counter")?
+        .ok_or_else(|| anyhow!("attach_request_dropped slot 0 missing"))?;
```
not-matthias: nit: formatting looks a bit off here

crates/memtrack/src/ebpf/attach_worker.rs
```diff
@@ -232,8 +232,40 @@ impl Worker {
...
+            // A mapping no path can reach costs coverage of that one library,
+            // which is not worth killing the run over.
+            if unreadable.len() == candidates.len() {
+                warn!(
+                    "no readable path for mapping {} in pid {}, allocator coverage may be \
+                     incomplete (tried {})",
+                    mapping.display,
+                    req.pid,
+                    unreadable.join(", ")
+                );
```
not-matthias, line 262: we should ensure this is not too verbose. i think the `unreadable` vec could be logged in debug instead.

not-matthias (approved):
LGTM!

crates/memtrack/src/ebpf/c/utils/event_helpers.h
GuillaumeLagrange: I've changed this to make use of a shared helper that returns both for call locations where we need both

PR #458, opened 2026-07-17
feat(memtrack): record per-benchmark pid in execution timestamps

GuillaumeLagrange:
Memory mode tracks the whole runner process tree, so between Start/StopProfiler the harness parent and sibling worker processes emit allocation events into every measured region. The CurrentBenchmark FIFO command already carries the pid that ran each benchmark; capture it alongside the uri in ExecutionTimestamps.bench_pid_by_ts (serde-default for artifacts written before this field) so the parser can scope events to the benchmarking process.


src/executor/shared/fifo.rs
```diff
@@ -206,7 +207,9 @@ impl RunnerFifo {
                 // Fall through to shared implementation for standard commands
                 match &cmd {
                     FifoCommand::CurrentBenchmark { pid, uri } => {
-                        bench_order_by_timestamp.push((get_current_time(), uri.to_string()));
+                        let ts = get_current_time();
```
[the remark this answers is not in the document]
GuillaumeLagrange: this is a known behavior that is not ideal, and not a regression compared to what exists for now, let's leave it at that

crates/memtrack/src/ebpf/c/memtrack.bpf.c
```diff
@@ -87,6 +112,29 @@ int tracepoint_sched_fork(struct trace_event_raw_sched_process_fork* ctx) {
         bpf_map_update_elem(&pids_ppid, &child_pid, &parent_pid, BPF_ANY);
 
         // bpf_printk("auto-tracking child process: child_pid=%u", child_pid);
+
+        /* Emit a fork event so userspace can rebuild the process tree and scope
+         * allocations to a process plus its descendants. Gated on is_enabled()
+         * like allocation events, so only forks inside a measured region reach
+         * the ring buffer. header.pid carries the child: current here is the
+         * parent, so we cannot use the SUBMIT_EVENT macro. */
+        if (is_enabled()) {
```
not-matthias, line 121: I dont think we should do this type of filtering here, because it's still useful to be able to track the full execution tree. We have the timestamp so we can do the processing in the backend

not-matthias (approved):
LGTM overall

crates/memtrack/src/ebpf/c/memtrack.bpf.c
not-matthias: e.g. for RSS tracking I also needed to track fork/exec/exit even when memtrack isnt enabled: [`8ef53f4` (#453)](https://github.com/CodSpeedHQ/codspeed/pull/453/changes/8ef53f4e60f308c22146159f23d689fff7ca3c2a)

PR #461, opened 2026-07-20
feat: enable cycle estimation and allocation exclusion by default

not-matthias:
## Summary

Graduate `--cycle-estimation` and `--exclude-allocations` out of the experimental flag set so both simulation behaviors are enabled for everyone by default.

- Both are now normal (non-experimental) options with `default_value_t = true`.
- They remain toggleable escape hatches: `--cycle-estimation=false` / `--exclude-allocations=false` (or env `CODSPEED_CYCLE_ESTIMATION`, `CODSPEED_EXCLUDE_ALLOCATIONS`) disable them.
- Dropped from the `Experimental` help heading and no longer trigger the experimental-flags warning.
- `--experimental-fair-sched` is intentionally left untouched (still experimental, default off).

## Changes

| File | Change |
|---|---|
| `src/cli/shared.rs` | Add `cycle_estimation` / `exclude_allocations` toggleable bools (default true, `require_equals`) to `ExecAndRunSharedArgs` |
| `src/cli/experimental.rs` | Remove both from `ExperimentalArgs`; only `experimental_fair_sched` remains |
| `src/cli/run/mod.rs`, `src/cli/exec/mod.rs` | Read the new shared fields |
| `src/executor/config.rs` | `OrchestratorConfig::test()` defaults both true |

## Behavior note

`Runner.exclude_allocations` in upload metadata now serializes `true` on every simulation run (previously omitted via `skip_serializing_if`). This is the intended signal, just always present now.

## Verification

- `cargo build`, `cargo fmt --check`, `cargo clippy --all-targets` clean.
- Verified via the built binary: `--cycle-estimation`, `--cycle-estimation=false`, and `--exclude-allocations=false` all parse; help lists both as normal options.
- Pre-existing `Unsupported system` valgrind-executor test failures are environment-gated (unsupported valgrind host) and unrelated to this change.

### DIFF
--- a/src/cli/experimental.rs
+++ b/src/cli/experimental.rs
@@ -17,22 +17,12 @@ pub struct ExperimentalArgs {
     )]
     pub experimental_fair_sched: bool,
 
-    /// Enable Valgrind cycle estimation (--cycle-estimation) in simulation mode.
-    #[arg(
-        long,
-        default_value_t = false,
-        help_heading = "Experimental",
-        env = "CODSPEED_EXPERIMENTAL_CYCLE_ESTIMATION"
-    )]
+    /// Deprecated: cycle estimation is enabled by default and this flag has no effect.
+    #[arg(long, hide = true, env = "CODSPEED_EXPERIMENTAL_CYCLE_ESTIMATION")]
     pub experimental_cycle_estimation: bool,
 
-    /// Exclude memory allocation time from simulation results.
-    #[arg(
-        long,
-        default_value_t = false,
-        help_heading = "Experimental",
-        env = "CODSPEED_EXPERIMENTAL_EXCLUDE_ALLOCATIONS"
-    )]
+    /// Deprecated: allocation exclusion is enabled by default and this flag has no effect.
+    #[arg(long, hide = true, env = "CODSPEED_EXPERIMENTAL_EXCLUDE_ALLOCATIONS")]
     pub experimental_exclude_allocations: bool,
 }
 
@@ -42,12 +32,6 @@ impl ExperimentalArgs {
         let mut flags = Vec::new();
         if self.experimental_fair_sched {
             flags.push("--experimental-fair-sched");
-        }
-        if self.experimental_cycle_estimation {
-            flags.push("--experimental-cycle-estimation");
-        }
-        if self.experimental_exclude_allocations {
-            flags.push("--experimental-exclude-allocations");
         }
         flags
     }
@@ -73,5 +57,34 @@ impl ExperimentalArgs {
             flag_list,
             style("https://github.com/CodSpeedHQ/codspeed/issues").underlined(),
         );
+    }
+
+    /// Warns about deprecated flags that were graduated to default-on options and
+    /// no longer have any effect.
+    pub fn warn_if_deprecated(&self) {
+        let deprecated = [
+            (
+                self.experimental_cycle_estimation,
+                "--experimental-cycle-estimation",
+                "cycle estimation",
+                "--cycle-estimation",
+            ),
+            (
+                self.experimental_exclude_allocations,
+                "--experimental-exclude-allocations",
+                "allocation exclusion",
+                "--exclude-allocations",
+            ),
+        ];
+
+        for (_, flag, feature, new_flag) in deprecated.iter().filter(|(set, ..)| *set) {
+            eprintln!(
+                "  {} {} has no effect: {} is now controlled by {} (defaults to true).",
+                style(Icon::Warning.to_string()).yellow(),
+                style(*flag).bold(),
+                feature,
+                style(*new_flag).bold(),
+            );
+        }
     }
 }
--- a/src/cli/shared.rs
+++ b/src/cli/shared.rs
@@ -113,6 +113,24 @@ pub struct ExecAndRunSharedArgs {
     #[arg(long)]
     pub base: Option<String>,
 
+    /// Enable Valgrind cycle estimation (--cycle-estimation) in simulation mode.
+    #[arg(
+        long,
+        env = "CODSPEED_CYCLE_ESTIMATION",
+        default_value_t = true,
+        action = clap::ArgAction::Set
+    )]
+    pub cycle_estimation: bool,
+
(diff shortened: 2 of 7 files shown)


src/cli/experimental.rs
```diff
@@ -16,24 +16,6 @@ pub struct ExperimentalArgs {
...
     pub experimental_fair_sched: bool,
-
-    /// Enable Valgrind cycle estimation (--cycle-estimation) in simulation mode.
-    #[arg(
-        long,
-        default_value_t = false,
-        help_heading = "Experimental",
-        env = "CODSPEED_EXPERIMENTAL_CYCLE_ESTIMATION"
-    )]
-    pub experimental_cycle_estimation: bool,
```
GuillaumeLagrange, line 27: we should keep them with a warning for a few release cycle (remove them in a few months)

the warning would say that this flag has no effect and the feature has been enabled by default + new name of the flag to disable

src/cli/shared.rs
```diff
@@ -113,6 +113,30 @@ pub struct ExecAndRunSharedArgs {
...
 
+    /// Enable Valgrind cycle estimation (--cycle-estimation) in simulation mode.
+    #[arg(
+        long,
+        env = "CODSPEED_CYCLE_ESTIMATION",
+        default_value_t = true,
+        default_missing_value = "true",
+        num_args = 0..=1,
+        require_equals = true,
+        action = clap::ArgAction::Set
```
GuillaumeLagrange, line 121: these are a bit too explicit, we can align this on other bool clap flags

GuillaumeLagrange (approved):
olgtm, make sure to not merge before we minor release

src/cli/experimental.rs
not-matthias: > + new name of the flag to disable

Do we need this? Since the flag right now is opt-in only so if someone was using it they likely wouldn't want to disable it, wdyt?

not-matthias: Nvm, thinking about it, it makes more sense to have it like this.

GuillaumeLagrange: Actually the new name of the flag to disable makes no sense indeed since it was opt in, but we should print the new flag name, and says it defaults to true now

src/cli/experimental.rs
```diff
@@ -17,22 +17,12 @@ pub struct ExperimentalArgs {
...
-    /// Exclude memory allocation time from simulation results.
-    #[arg(
-        long,
-        default_value_t = false,
-        help_heading = "Experimental",
-        env = "CODSPEED_EXPERIMENTAL_EXCLUDE_ALLOCATIONS"
-    )]
+    /// Deprecated: allocation exclusion is enabled by default and this flag has no effect.
+    #[arg(long, hide = true, env = "CODSPEED_EXPERIMENTAL_EXCLUDE_ALLOCATIONS")]
     pub experimental_exclude_allocations: bool,
```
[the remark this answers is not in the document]
not-matthias: This is fine, experimental flags are mostly unused by users. And since the default value was false, the chance that someone was explicitly disabling it is almost 0.

src/cli/shared.rs
```diff
@@ -113,6 +113,30 @@ pub struct ExecAndRunSharedArgs {
...
+    )]
+    pub cycle_estimation: bool,
+
+    /// Exclude memory allocation time from simulation results.
+    #[arg(
+        long,
+        env = "CODSPEED_EXCLUDE_ALLOCATIONS",
+        default_value_t = true,
```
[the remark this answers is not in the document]
not-matthias: False positive. The metadata hash isn't a cross-run identity/comparability key — it's used only for tokenless log-validation (`tokenlessValidation.ts`): the backend recomputes it from the same uploaded payload within a single run and matches it against the CI logs to prove the run is genuine. It's never compared against a stored hash from another run, so there's no old/new hash-semantics mixing and no migration needed.

Cross-run comparability is handled separately by `excludeAllocationsRule` (`compatibility/rules.ts`), which emits a *neutral* "not directly comparable" incompatibility when base and head differ on the flag (covered by tests). So the base(off)→head(on) transition after a runner upgrade degrades gracefully to a clear note, not a broken baseline.

The backend has modeled `excludeAllocations` as an optional field since the experimental phase (`Runner.ts`, full parse pipeline), so the payload schema — and `LATEST_UPLOAD_METADATA_VERSION` — is unchanged. `skip_serializing_if` keeps the opt-out path (`--exclude-allocations=false`) byte-identical to legacy runs.

src/cli/shared.rs
```diff
@@ -113,6 +113,24 @@ pub struct ExecAndRunSharedArgs {
...
+        action = clap::ArgAction::Set
```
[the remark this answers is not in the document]
not-matthias: Checked the backend — this is a false positive.

`getMetadataHash` has exactly one functional call site: `services/upload/tokenlessValidation.ts`. It's used only to prove a tokenless run is genuine by matching the hash against the CI job's public logs. The hash is **never persisted** (`RunPart` stores `rawMetadata`, not the hash) and **never compared across runs**.

Crucially it's self-consistent within a single run — the backend computes it from the same payload the runner uploaded and logged (comment in `tokenlessValidation.ts`: *"runner hashes `serde_json::to_string(metadata)`, we mirror that by hashing `JSON.stringify(JSON.parse(rawMetadata))`"*). So `excludeAllocations: true` being present in both sides just matches; there's no legacy hash to collide with.

So "the same benchmark invocation can get a different run hash after this release" is true but inconsequential: the hash is recomputed fresh each run and only checked against that run's own logs. The KB change-checklist note is about *adding/removing a field* (schema change) with an *unconditional* serialization — neither applies here: the field already exists, and `skip_serializing_if` is exactly the opt-in pattern that note endorses. This PR only flips the default, so no `LATEST_UPLOAD_METADATA_VERSION` bump is needed.

Cross-run comparability (the thing that actually matters when base/head differ) is handled separately by `excludeAllocationsRule` in `compatibility/rules.ts`, which emits a neutral "not directly comparable" note.

not-matthias:
@greptileai review

PR #469, opened 2026-07-24
feat(memtrack): support disabling allocator tracking

not-matthias:
## What

`memtrack track --track-allocators` (default on, env `CODSPEED_TRACK_ALLOCATORS`). When disabled, memtrack skips the allocator uprobe machinery and only emits coarse `mmap`/`munmap`/`brk` events.

## Why

Allocation-heavy programs (e.g. a Rust build) generate an overwhelming number of `malloc`/`free` events, and the per-allocation uprobes slow the target significantly. This trades allocation granularity for lower overhead while still collecting RSS-relevant memory events.

## How

- `mmap`/`munmap`/`brk` syscall tracepoints are now always attached in a memory run.
- With the flag off, the exec watcher and allocator attach worker are never started.
- No runner CLI flag: the env var is inherited by the memtrack subprocess, keeping the runner decoupled from the installed memtrack version.

### DIFF
--- a/crates/memtrack/src/ebpf/tracker.rs
+++ b/crates/memtrack/src/ebpf/tracker.rs
@@ -8,6 +8,30 @@ use std::os::unix::process::CommandExt;
 use std::process::Command;
 use std::sync::Arc;
 use std::sync::mpsc;
+use typed_builder::TypedBuilder;
+
+#[derive(Debug, Clone, Copy, TypedBuilder)]
+pub struct TrackerOptions {
+    /// Attach allocator uprobes (malloc/free/calloc/...) through the
+    /// exec-mapping watcher.
+    #[builder(default = true)]
+    pub allocators: bool,
+    /// Reconstruct per-process RSS from the folio rmap fentry hooks.
+    #[builder(default = false)]
+    pub rmap: bool,
+}
+
+impl TrackerOptions {
+    fn from_env() -> Self {
+        Self::builder()
+            .allocators(!matches!(
+                std::env::var("CODSPEED_MEMTRACK_TRACK_ALLOCATORS").as_deref(),
+                Ok("0") | Ok("false")
+            ))
+            .rmap(std::env::var("CODSPEED_MEMTRACK_TRACK_RMAP").is_ok_and(|v| v == "1"))
+            .build()
+    }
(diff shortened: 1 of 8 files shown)


crates/memtrack/src/main.rs
```diff
@@ -30,6 +30,17 @@ enum Commands {
...
         ipc_server: Option<String>,
+
+        /// Track individual heap allocations (malloc/free/calloc/...). Disable to
+        /// reduce overhead on allocation-heavy programs; mmap/brk events are still
+        /// collected either way.
+        #[arg(
+            long,
+            env = "CODSPEED_TRACK_ALLOCATORS",
+            default_value_t = true,
+            action = clap::ArgAction::Set
```
GuillaumeLagrange: not sure we need to explicitly set `action = clap::ArgAction::Set` here

GuillaumeLagrange (commented):
I'm not sure what's really the usecase of this? olgtm but I'm not sure we've discussed

crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -8,6 +8,33 @@ use std::os::unix::process::CommandExt;
 use std::process::Command;
 use std::sync::Arc;
 use std::sync::mpsc;
+use typed_builder::TypedBuilder;
+
+/// Which probes a tracker attaches. [`Tracker::new`] resolves these from the
+/// environment, [`Tracker::with_options`] takes them directly.
```
GuillaumeLagrange: Kind of superfluous comment here

crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -8,6 +8,33 @@ use std::os::unix::process::CommandExt;
...
+#[derive(Debug, Clone, Copy, TypedBuilder)]
+pub struct TrackerOptions {
+    /// Attach allocator uprobes (malloc/free/calloc/...) through the
+    /// exec-mapping watcher. The coarse mmap/munmap/brk tracepoints fire either
```
GuillaumeLagrange: Last sentence is not really necessary

GuillaumeLagrange (approved):
olgtm

PR #470, opened 2026-07-24
fix(memtrack): track successful posix_memalign allocations

not-matthias:
## Problem

`posix_memalign` was attached to the shared `memalign` uprobe handler (`UPROBE_ARG_RET`), which assumes the standard allocator ABI. But `posix_memalign(void** memptr, size_t alignment, size_t size)` is inverted:

- Returns `int 0` on **success** (nonzero errno on failure), not a pointer.
- Delivers the allocated pointer through the `memptr` out-parameter, not the return register.
- Size lives in `PARM3` (`PARM2` is the alignment).

The shared handler gated success on a non-NULL return and used the return register as the address. Since successful `posix_memalign` returns `0`, every successful allocation was treated as a failure and **silently dropped**.

## Fix

Dedicated uprobe/uretprobe pair for `posix_memalign`:

- uprobe stashes `memptr` (PARM1) and `size` (PARM3) keyed by tid.
- uretprobe gates on `RC == 0`, reads the address back from `*memptr` via `bpf_probe_read_user`, then submits an aligned-alloc event.

## Test

Added `posix_memalign_test` to the existing rstest table in `tests/c_tests.rs` with a fixture + snapshot confirming the allocation is now tracked.

### DIFF
--- a/crates/memtrack/src/ebpf/c/allocator.h
+++ b/crates/memtrack/src/ebpf/c/allocator.h
@@ -90,6 +90,57 @@ UPROBE_ARG_RET(aligned_alloc, PT_REGS_PARM2(ctx),
 
 UPROBE_ARG_RET(memalign, PT_REGS_PARM2(ctx), { return submit_aligned_alloc_event(arg0, ret_val); })
 
+/*
+ * posix_memalign(void** memptr, size_t alignment, size_t size)
+ *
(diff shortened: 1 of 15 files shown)


GuillaumeLagrange (approved):
LGTM, I'm guessing you've double checked that we do not double dip and cannot count allocations twice because of this change

PR #490, opened 2026-07-29
chore: make --exclude-allocations nicer to use

not-matthias:
Before you had to pass =true or =false instead of just using --exclude-allocations


not-matthias:
@greptileai review

not-matthias:
@greptileai review again

GuillaumeLagrange (approved):

PR #506, opened 2026-08-18
refactor(logger): scope rolling buffer display behind provider 

GuillaumeLagrange:
Gets rid of the ugly unconditionnal bypass that relies on *IS_TTY

@fargito FYI, I will only merge this once you're done with the circle ci changes, I'll handle the conflicts, it's a cleaner solution to what we discussed this morning


fargito (commented):
Looks cleaner indeed

fargito:
@GuillaumeLagrange now that the CircleCI stuff is released, you can rebase

PR #507, opened 2026-08-18
feat(memory): apply kernel memory tunables before memory-mode runs

not-matthias:
Benchmark repos running `codspeed --mode memory` currently need a hand-written CI step to stabilise kernel memory behaviour before measuring: disable THP, zero `vm.compaction_proactiveness` / `vm.swappiness` / `kernel.numa_balancing`, `swapoff -a`, and drop the page cache before each suite.

The runner now does this itself at the start of every memory-mode execution.

## Behaviour

- Applied only when a CI run environment is detected, so developer machines are not silently mutated. Skipped entirely when privileges cannot be elevated without a password prompt.
- Best effort: a knob that cannot be applied is a warning, never fatal.
- `MemoryTunables::apply()` returns a guard holding the previous value of every knob it changed, and restores it on drop — including on the error paths, where `teardown()` never runs. Only knobs that were not already at the target are captured, and THP is restored to the exact mode it held (`madvise` stays `madvise`).
- `swapoff -a` is guarded: skipped on zram devices (it permanently resets their `disksize`) and whenever the swapped pages would not fit in `MemAvailable`. Restoring re-enables the recorded entries one by one, since `swapon -a` would miss a swap file absent from `/etc/fstab`.
- The page cache drop has no previous state to restore; the node is a write-only trigger.
- Applied in `run()` rather than `setup()`, so the page cache is dropped before every suite and `--skip-setup` does not bypass it. The sysctl/THP writes are no-ops on later suites since `ensure_sysctl` compares before writing.

## Commits

1. `refactor(executor): move linux_sysctl into executor helpers` — pure move, the sysctl primitive is no longer walltime-specific.
2. `feat(memory): apply kernel memory tunables before memory-mode runs`

## Verification

- `cargo clippy --all-targets` clean, `cargo test --lib` (pre-existing failures on this host are the sudo-requiring tests, which write `kernel.kptr_restrict` and are identical on `main`).
- Local non-CI run logs `Not running in CI, skipping kernel memory tunables`; host THP unchanged.
- Privileged container: THP `enabled`/`defrag` become `[never]`, `vm.compaction_proactiveness = 0`, `vm.swappiness = 0`, and the swap guard correctly logged `Leaving swap enabled: swapped pages do not fit in available memory`.


src/executor/memory/tunables.rs
```diff
@@ -0,0 +1,328 @@
...
+            );
+            return Self::default();
+        }
+
+        start_group!("Applying kernel memory tunables");
+        let tunables = Self {
+            thp: Self::set_thp("never"),
+            sysctls: Self::set_sysctls(0),
+            swap: Self::set_swap(false, &[]),
+        };
```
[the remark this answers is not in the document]
GuillaumeLagrange: Not an issue, any decent CI runner will either virtualize or have other layers to make sure anything a CI job does is not persistent, best effort here is fine IMO

src/executor/memory/tunables.rs
```diff
@@ -0,0 +1,324 @@
...
+        Self::drop_page_cache();
+        end_group!();
+
+        tunables
+    }
+
+    /// Drops the page cache. Nothing to restore: the node is a write-only
+    /// trigger and the kernel refills the cache on demand.
+    fn drop_page_cache() {
+        nix::unistd::sync();
```
GuillaumeLagrange, line 60: Why is this call here? Could benefit from a comment

src/executor/memory/tunables.rs
```diff
@@ -0,0 +1,324 @@
...
+    swap: Vec<String>,
+}
+
+impl MemoryTunables {
+    /// Applies the knobs on a best-effort basis: a knob that cannot be set is
+    /// warned about, never fatal.
+    pub fn apply() -> Self {
+        if !crate::run_environment::is_ci_environment() {
+            debug!("Not running in CI, skipping kernel memory tunables");
+            return Self::default();
```
GuillaumeLagrange: I think it's better to return an option here rather than relying on the default satisfies the condition of the drop guard. This way if we dont do anything, we are SURE we'll not run any command when dropping the tunables guard

src/executor/memory/tunables.rs
```diff
@@ -0,0 +1,324 @@
+//! Kernel knobs that stabilise memory measurements: transparent huge pages,
+//! compaction/swap/NUMA-balancing sysctls, swap and the page cache.
+//!
+//! [`MemoryTunables`] captures the previous value of every knob it changes and
+//! restores it on drop, so a host that only looks like CI — `CI=true` inside a
+//! container sharing the host's non-namespaced knobs, say — is left as it was.
```
GuillaumeLagrange: Could we have some traceability and documentation about these tuneables?

Ideally, each should be individually tested for its variance effect. This is to make sure we are not vibe-disabling swap because opus/fable felt like it was a good idea.

src/executor/memory/tunables.rs
```diff
@@ -0,0 +1,324 @@
...
+    sysctls: Vec<(&'static str, i64)>,
+    /// Swap entries that were active before `swapoff -a`.
+    swap: Vec<String>,
+}
+
+impl MemoryTunables {
+    /// Applies the knobs on a best-effort basis: a knob that cannot be set is
+    /// warned about, never fatal.
+    pub fn apply() -> Self {
+        if !crate::run_environment::is_ci_environment() {
```
GuillaumeLagrange: We are not doing this kind of guard for walltime, could we unify behavior? This feels very ad-hoc for now

not-matthias: Fair, removed it for now. Since it's only temporary it should be fine to run locally. I'll also swap the walltime tunables to be only temporary

src/executor/memory/tunables.rs
not-matthias: Agreed. I reduced this to the two controls; we can add further controls later.

Enabled settings:
- `transparent_hugepage/enabled=never`: prevents transparent huge pages from backing sparse anonymous mappings with a 2 MiB page, which can make the RSS peak depend on page-promotion timing (observed in experiments)
- `sync; vm.drop_caches=3`: gives each run the same initial clean page-cache and reclaimable-slab state. drop_caches is a one-shot reset, not a persistent policy.

I removed THP defrag because it is redundant once the normal THP allocation path is disabled, and removed proactive compaction, swapoff, swappiness, and NUMA balancing because they aren't really needed.

src/executor/helpers/linux_sysctl.rs
art049, line 1: Is it possible to gate this whole mod behind the feature flag but not every single line, because now we end up gating almost every single line with `#[cfg(target_os = "linux")]` 😅

art049 (changes requested):
I know it was the case before but it hurts readability a bit

not-matthias:
@greptileai review again

art049 (approved):

PR #522, opened 2026-08-28
feat(memtrack): capture allocation stacks in eBPF

not-matthias:
Adds raw allocation-stack capture to memtrack and ships everything needed for off-box unwinding.

At every allocator entry the eBPF program copies the caller's user stack (chunked, budget-capped, FNV-1a-hashed) and walks frame pointers in-kernel. Definitions are deduplicated by hash in-kernel: one `StackDefinition` event (raw bytes, registers, FP chain) per distinct stack, referenced by a `stack_hash` on allocation events. Everything flows through the ordinary `MemtrackArtifact` event stream; unwinding and symbolication happen server-side later.

After a capture-enabled run, the runner decodes the artifact and reuses the perf walltime machinery to dump per-module symbols, unwind data, and debug info into the profile folder, plus a `memory_metadata.json` with per-pid process names.

Refs COD-3222


crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -95,31 +123,60 @@ impl Tracker {
...
+                let mut bpf = self.bpf.lock();
+                bpf.add_tracked_pid(pid)?;
+                let stack_poller = capture_stacks
+                    .then(|| bpf.poll_stacks(10, tx.clone()))
+                    .transpose()?;
+                (bpf.poll_events_with_channel(10, tx.clone())?, stack_poller)
+            };
+            let perf_mapping_poller = capture_stacks
+                .then(|| PerfMappingPoller::start(pid, tx, self.mapping_lost.clone()))
+                .transpose()?;
```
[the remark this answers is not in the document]
GuillaumeLagrange: yet another argument for getting read of the perf token

crates/runner-shared/src/artifacts/memtrack/mod.rs
```diff
@@ -91,6 +102,41 @@ pub enum MemtrackEventKind {
...
+        ino: u64,
+        file_offset: u64,
+        len: u64,
+    },
+
+    Stack {
+        // Box keeps the MemtrackEventKind enum small across millions of events.
+        #[serde(flatten)]
+        record: Box<StackRecord>,
+    },
```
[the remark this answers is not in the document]
GuillaumeLagrange: Not an issue, we update the readers before the writers my dude

crates/memtrack/src/ebpf/events.rs
```diff
@@ -111,6 +122,74 @@ pub fn parse_event(data: &[u8]) -> Option<MemtrackEvent> {
...
+        Ok(None) => return Vec::new(),
+        Err(error) => {
+            warn!("Failed to read frame-pointer chain for stackid {stackid}: {error}");
+            return Vec::new();
+        }
+    };
+
+    // The map value is a fixed-depth array zero-padded past the last frame.
+    value
+        .chunks_exact(8)
```
GuillaumeLagrange, line 172: maybe overkill, but we should probable panic here if we're not unwinding a 64bit stack?

crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -20,9 +26,13 @@ pub struct TrackerOptions {
...
     pub physical: bool,
-    /// Uprobe attach mechanism. `None` detects it from BPF token availability.
-    #[builder(default, setter(strip_option))]
-    pub variant: Option<BpfVariant>,
+    /// Capture allocation call stacks, adding per-allocation stack walking and
+    /// raw stack copying.
+    #[builder(default = false)]
+    pub stack_capture: bool,
+    /// Maximum bytes of user stack to copy per captured call stack.
+    #[builder(default = 8192)]
```
GuillaumeLagrange, line 34: We'll most likely have to revisit this when doing python 🥲

crates/memtrack/src/ebpf/stacks.rs
```diff
@@ -0,0 +1,35 @@
...
+use crate::prelude::*;
+
+#[derive(Debug, Clone, Copy, Default, serde::Serialize)]
+pub struct StackCaptureStats {
+    pub copy_failed: u64,
+    pub hash_map_full: u64,
+    pub stackid_failed: u64,
+    pub truncated: u64,
+    pub ring_full: u64,
+}
```
GuillaumeLagrange: Should be renamed `StackCaptureFailureStats`

crates/memtrack/src/perf_mappings.rs
```diff
@@ -0,0 +1,437 @@
+use crate::prelude::*;
```
GuillaumeLagrange, line 1: Looks like at least some parts of the code could be shared with the runner itself couldn't it?

src/executor/memory/module_artifacts.rs
```diff
@@ -0,0 +1,669 @@
...
+        };
+
+        let process_loaded_module = loaded_module
+            .process_loaded_modules
+            .entry(mapping.pid)
+            .or_default();
+        process_loaded_module.symbols_load_bias = Some(load_bias);
+
+        if let Some(process_unwind_data) = process_unwind_data {
+            process_loaded_module.process_unwind_data = Some(process_unwind_data);
```
GuillaumeLagrange, line 230: Regarding the clanker's remark, it's a weakness we also share with the perf integration, we'll have to do timestamp aware mounts at some point but honnestly not worth it here is it?

crates/runner-shared/src/metadata.rs
```diff
@@ -11,35 +12,68 @@ use crate::fifo::MarkerType;
...
+    HashMap::<String, V>::deserialize(deserializer)?
+        .into_iter()
+        .map(|(key, value)| {
+            let pid = key
+                .parse::<pid_t>()
+                .map_err(|_| D::Error::custom(format!("invalid pid key: {key}")))?;
+            Ok((pid, value))
+        })
+        .collect()
+}
```
GuillaumeLagrange, line 38: It feels really weird that we need this when we didnt before?

GuillaumeLagrange (approved):
OLGTM, quite a big piece of work

crates/memtrack/src/ebpf/events.rs
GuillaumeLagrange: Or have the word length dynamically outputted by the data emitter and parsed here.
We're still far from embedded but may as well make the code portable 🤓

crates/runner-shared/src/metadata.rs
GuillaumeLagrange: We've been deserializing JSON into these structs for a while now

not-matthias: Yeah, I was also quite confused by this. But it's essentially a limitation of serde when using `#[serde(flatten)]` (see link in comment) which got introduced when fields got moved from `WalltimeMetdata` into `ModuleArtifacts` to have: 
```
    #[serde(flatten)]
    pub artifacts: ModuleArtifacts,
```

GuillaumeLagrange: Aight, all good then!

crates/memtrack/src/ebpf/tracker.rs
not-matthias: I know :(

crates/memtrack/src/perf_mappings.rs
not-matthias: Not that much, since it's mostly responsible for the emission/collection. But I just found the perf-event2 package which can simplify a lot of this

src/executor/memory/module_artifacts.rs
not-matthias: Agreed, keeping timestamp-aware placements out of this PR. Added a code comment locally documenting that only the last placement per (path, pid) is retained, as in perf, and referencing COD-1377

src/executor/memory/module_artifacts.rs
```diff
@@ -0,0 +1,671 @@
...
```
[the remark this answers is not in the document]
not-matthias: As mentioned in the code comment, we should work on this in the future as perf has the same limitation right now

crates/memtrack/src/ebpf/events.rs
not-matthias: I think it's a bit overkill to implement different pointer lengths. We'd need to make changes in a lot of other places (e.g. backend, instrument-hooks, etc.) and also validate that eBPF is available. 

So I just added a panic for now to have a bit of protection for this in the future.

crates/memtrack/src/perf_mappings.rs
```diff
@@ -0,0 +1,264 @@
...
+                [first] => RawData::Single(first),
+                [first, second] => RawData::Split(first, second),
+                _ => unreachable!("perf records contain one or two slices"),
+            };
+            match RecordType(record.ty()) {
+                RecordType::MMAP2 => {
+                    let raw =
+                        RawEventRecord::new(RecordType::MMAP2, record.misc(), body, parse_info);
+                    if let Some(event) = parse_mmap2(raw) {
+                        mappings.push(event);
```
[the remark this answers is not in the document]
not-matthias: Not really an issue since we're benching workloads taking a few minutes at most. we'll have at most tens of MB which is still not that big

crates/memtrack/src/ebpf/memtrack/fd_holder.rs
```diff
@@ -0,0 +1,401 @@
...
+            unsafe {
+                libc::close_range(lo, fd - 1, 0);
+            }
+        }
+        lo = fd.saturating_add(1);
+    }
+    // SAFETY: same as above.
+    unsafe {
+        libc::close_range(lo, u32::MAX, 0);
+    }
```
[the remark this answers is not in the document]
not-matthias: Since we need 5.11+, this should always be available (since it was introduced in v5.9)

PR #524, opened 2026-08-31
feat(memtrack): fail early when the kernel has no BTF

not-matthias:
## Problem

Memory mode fails on runners whose kernel is built without `CONFIG_DEBUG_INFO_BTF`.

libbpf resolves CO-RE relocations against the running kernel's own BTF, and the kernel resolves the attach target of every `fentry`/`tp_btf` program against it as well. Without `/sys/kernel/btf/vmlinux` the programs cannot load at all — and libbpf surfaces this as a bare `-ESRCH`, which says nothing about the actual cause.

## Change

Preflight check in `MemtrackBpf::with_variant`, where the load would otherwise fail:

```rust
pub struct KernelBtf;

impl KernelBtf {
    pub fn is_available() -> bool;
    pub fn ensure_available() -> Result<()>;
}
```

The error names the running kernel and the config option to look for:

```
Memory profiling is not supported on this runner: its kernel (7.2.0) was built
without BTF, so /sys/kernel/btf/vmlinux does not exist. Use a runner whose kernel
is built with CONFIG_DEBUG_INFO_BTF=y.
```

The first commit is a pure refactor extracting the `/proc/sys/kernel/osrelease` read out of `KernelVersion::current`, so the message can report the running release.

## Verification

- `cargo test --release -p memtrack` — 19 passed, remainder are the pre-existing root/eBPF-gated ignored tests.
- `cargo fmt --check`, `cargo clippy --all-targets` clean.
- Detection exercised on a BTF-having host: `is_available()` returns `true`, `ensure_available()` returns `Ok`, and the failure message renders with the real kernel release.

## Notes for review

`is_available` is not unit-testable as written because the path is a hardcoded `const`. If coverage is wanted, it can take the path as a parameter behind a private helper.

COD-3417

### DIFF
--- a/crates/memtrack/src/kernel.rs
+++ b/crates/memtrack/src/kernel.rs
@@ -1,6 +1,15 @@
 use crate::prelude::*;
 use std::fmt;
 
+/// The running kernel's full release, e.g. `6.12.8+`.
+fn kernel_release() -> Result<String> {
+    const OSRELEASE_PATH: &str = "/proc/sys/kernel/osrelease";
+
+    std::fs::read_to_string(OSRELEASE_PATH)
+        .map(|release| release.trim().to_owned())
+        .with_context(|| format!("Failed to read {OSRELEASE_PATH}"))
+}
+
 /// A kernel release, ordered by `(major, minor)`. The patch level is ignored:
 /// features are introduced in merge windows, never in a stable point release.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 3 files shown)


crates/memtrack/src/kernel.rs
```diff
@@ -46,6 +51,40 @@ impl fmt::Display for KernelVersion {
...
+    pub fn ensure_available() -> Result<()> {
+        if Self::is_available() {
+            return Ok(());
+        }
+
+        let release = kernel_release().unwrap_or_else(|_| "unknown".to_owned());
+        bail!(
+            "Memory profiling is not supported on this runner: its kernel ({release}) \
+             was built without BTF, so {} does not exist. Use a runner whose kernel is \
+             built with CONFIG_DEBUG_INFO_BTF=y.",
```
GuillaumeLagrange: I think we can be a bit more concise here, no need to expose the full check, just say that this kernal was built without BTF support and that's enough IMO

GuillaumeLagrange (approved):
ogltm

PR #525, opened 2026-09-03
fix: support distributions CodSpeed publishes no valgrind package for (Arch, rolling releases)

moha-bekh:
Refs COD-3072

The CLI aborted on Arch before doing anything: `SupportedOs::from_os` failed when `/etc/os-release` had no `VERSION_ID`, which rolling releases do not ship. Fixing that surfaced the next failure at setup, where `install_valgrind` went straight to the `valgrind-codspeed` deb we only publish for some Debian/Ubuntu versions.

## Changes

- A missing OS version is now an `Option` rather than an error (macOS keeps the hard error, the version is always there and we report it to the API).
- `ValgrindExecutor::setup()` matches on its own `support_level`: deb install, source build, or bail.
- On a host we publish no package for, the setup accepts a valgrind the user brings, offers to build one from source (`CODSPEED_VALGRIND_BUILD_FROM_SOURCE` skips the prompt), and otherwise says what to install by hand.
- The libc debug symbol probe is scoped to the apt path, where it belongs, and warns instead of blocking.
- `codspeed setup` no longer skips `RequiresManualInstallation` executors; a failing one warns and the others still run.

## Blocked on

`VALGRIND_CODSPEED_BRANCH` points at a feature branch until CodSpeedHQ/valgrind-codspeed#40 lands. Merging before that would have released runners cloning it.

## Test plan

- [x] fmt, clippy, `cargo test --lib`.
- [x] manual run on Arch Linux ARM rolling (no `VERSION_ID`): `status`, `setup` and `exec` in the three modes pass, `7c4b2aa^` still reproduces the failure, and the no-package path warns and carries on.
- [x] the source build end to end on the same host: clone, vendored Capstone, configure, `make`, `make install` under sudo, then `setup`, `status` and `exec` against the freshly built `valgrind-3.26.0.codspeed7`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/system/os.rs
```diff
@@ -22,14 +26,21 @@ impl SupportedOs {
...
     /// The OS version is read from `sysinfo::System::os_version()`.
     pub fn from_os(os: &str) -> Result<Self> {
-        let os_version = System::os_version().ok_or(anyhow!("Failed to get OS version"))?;
         match os {
             "linux" => {
                 let os_id = System::distribution_id();
+                // Rolling release distributions (Arch, Gentoo, ...) do not expose a `VERSION_ID`
+                // in `/etc/os-release`, so `sysinfo` reports no version for them. This is not
+                // fatal: the version only matters for the distributions we ship packages for,
+                // which all expose one.
```
GuillaumeLagrange, line 35: Rephrase to to be one short sentence that says rolling distribs do not expose a version id

src/system/os.rs
```diff
@@ -22,14 +26,21 @@ impl SupportedOs {
...
+                let os_version = System::os_version().unwrap_or_else(|| {
+                    debug!("No OS version reported for distribution {os_id}");
+                    UNKNOWN_OS_VERSION.to_string()
+                });
```
GuillaumeLagrange, line 39: Nitpicky: wouldnt it be better to make the `version` field in other linux distro an option rather than hardocding an `unknown` string value. As far as I see, it's only used for display purposes

src/executor/valgrind/setup.rs
```diff
@@ -225,30 +226,97 @@ fn has_debug_symbols(binary: &Path) -> bool {
...
+fn has_libc_debug_symbols(system_info: &SystemInfo) -> bool {
+    system_libc_path(system_info).is_some_and(|libc| has_debug_symbols(&libc))
+}
+
+/// Warn, without failing, when valgrind will run against a libc it has no debug
+/// symbols for.
+///
+/// The symbols sharpen valgrind's output but are not needed to run it, so a missing
+/// `libc6-dbg` must not reject an installation we did not package ourselves. Scoped to
+/// apt-based systems, the only layout [`system_libc_path`] knows.
```
GuillaumeLagrange, line 253: slop comment

src/executor/valgrind/setup.rs
```diff
@@ -225,30 +226,97 @@ fn has_debug_symbols(binary: &Path) -> bool {
...
+    )
+}
 
+/// Whether the system libc has a resolvable separate debug file, as the `libc6-dbg`
+/// package provides.
+///
+/// Probed by file rather than through dpkg, because the setup cache restores package
+/// files onto the root filesystem without touching dpkg's database. Only meaningful
+/// where [`system_libc_path`] resolves, so callers must already know they are on the
+/// Debian multiarch layout.
```
GuillaumeLagrange, line 243: slop comment

src/executor/valgrind/setup.rs
```diff
@@ -225,30 +226,97 @@ fn has_debug_symbols(binary: &Path) -> bool {
...
 }
 
 pub async fn install_valgrind(
     system_info: &SystemInfo,
     setup_cache_dir: Option<&Path>,
 ) -> Result<()> {
+    // On distributions we do not publish a valgrind-codspeed package for (rolling releases,
+    // non-apt systems, ...), there is nothing to install automatically: the user brings their own
+    // build. Accept that installation instead of failing on the package we cannot provide.
+    if !is_codspeed_valgrind_installation_supported(system_info) {
```
GuillaumeLagrange, line 275: I think this logic belongs in the valgrind executor setup function.
We could call the support level helper and match against it instead of re-implementing the logic here.

Then in this file, we would get an install valgrind from apt, and a `try_install_from_source` depending on the support level.

src/executor/valgrind/setup.rs
```diff
@@ -225,30 +226,97 @@ fn has_debug_symbols(binary: &Path) -> bool {
...
+        if is_valgrind_installed() {
+            debug!(
+                "Using the valgrind installation already present on {}",
+                system_info.os
+            );
+            warn_on_missing_libc_debug_symbols(system_info);
```
GuillaumeLagrange, line 281: this is nice!

src/executor/valgrind/build_from_source.rs
```diff
@@ -0,0 +1,286 @@
...
+        }
+        "opensuse-tumbleweed" | "opensuse-leap" | "sles" => {
+            "sudo zypper install -y -t pattern devel_basis git"
+        }
+        "alpine" => "sudo apk add build-base autoconf automake git",
+        "ubuntu" | "debian" | "linuxmint" | "pop" | "raspbian" => {
+            "sudo apt-get install -y build-essential autoconf automake git"
+        }
+        _ => return None,
+    };
```
GuillaumeLagrange, line 95: I like the idea, but IMO we should just list the dependencies that were not found, and not bother with having to maintain this list, we already do a lot for this edge case, may as well keep the code stupid

src/executor/valgrind/build_from_source.rs
```diff
@@ -0,0 +1,286 @@
...
+        bail!("`{command_line}` failed with {status}");
+    }
+
+    Ok(())
+}
+
+/// Clone the sources from scratch, so that a partial or outdated checkout left
+/// over by a previous attempt never leaks into the build.
+async fn clone_sources() -> Result<PathBuf> {
+    let source_dir = env::temp_dir().join(SOURCE_DIR_NAME);
```
GuillaumeLagrange, line 131: use tempfile, a rust crate we already use in this project, you wont have to bother about whether the file exists or not

src/executor/valgrind/build_from_source.rs
```diff
@@ -0,0 +1,286 @@
...
+/// Environment variable that answers [`is_wanted`] without asking, for CI and any
+/// other unattended run that wants the opposite of the default.
+const BUILD_FROM_SOURCE_ENV: &str = "CODSPEED_VALGRIND_BUILD_FROM_SOURCE";
+
+/// Branch of the valgrind-codspeed repository to build from.
+// TODO: switch back to `main` once the self-contained build script has landed there.
+const VALGRIND_CODSPEED_BRANCH: &str = "cod-3465-create-a-self-contained-valgrind-build-script";
+
+/// Directory name, under the system temporary directory, the sources are cloned into.
+const SOURCE_DIR_NAME: &str = "valgrind-codspeed-src";
```
GuillaumeLagrange, line 36: not needed with `tempdir`

src/executor/valgrind/build_from_source.rs
```diff
@@ -0,0 +1,286 @@
...
+         with sudo. Declining leaves the installation to you, see \
+         https://github.com/CodSpeedHQ/valgrind-codspeed"
+    );
+    eprint!("\nBuild valgrind-codspeed from source now? [Y/n] ");
+
+    let line = Term::stderr().read_line().unwrap_or_default();
+    let answer = line.trim();
+
+    let accepted =
+        answer.is_empty() || answer.eq_ignore_ascii_case("y") || answer.eq_ignore_ascii_case("yes");
```
GuillaumeLagrange, line 233: Naive question, so maybe it's legit: why is an empty answer considered accepted?

src/executor/valgrind/build_from_source.rs
```diff
@@ -0,0 +1,286 @@
...
+    #[test]
+    fn is_wanted_honours_an_explicit_env_var() {
+        temp_env::with_var(BUILD_FROM_SOURCE_ENV, Some("true"), || {
+            assert!(is_wanted());
+        });
+        temp_env::with_var(BUILD_FROM_SOURCE_ENV, Some("false"), || {
+            assert!(!is_wanted());
+        });
+    }
+}
```
GuillaumeLagrange, line 286: these tests are quite useless, you cna remove them

GuillaumeLagrange (changes requested):
olgtm, nothing that big to change!

src/executor/valgrind/build_from_source.rs
moha-bekh: It's juste the default behavior, [Y/n] mecanism

moha-bekh:
@GuillaumeLagrange ready for another pass — all threads addressed, and the manual Arch run is done (`status`, `setup`, `exec` in the three modes). Still blocked on CodSpeedHQ/valgrind-codspeed#40.

src/executor/valgrind/setup.rs
```diff
@@ -225,30 +226,89 @@ fn has_debug_symbols(binary: &Path) -> bool {
...
+    bail!(
+        "valgrind-codspeed {} or higher is required and could not be installed automatically. \
+        Install it manually, see https://github.com/CodSpeedHQ/valgrind-codspeed",
+        VALGRIND_CODSPEED_VERSION_STRING.as_str()
+    );
 }
 
-pub async fn install_valgrind(
+/// Install the valgrind-codspeed package we publish for this system, from apt.
+pub(super) async fn install_valgrind(
```
GuillaumeLagrange: let's rename this function to `install_valgrind_from_package` in opposition to `try_install_from_source`

GuillaumeLagrange (approved):
olgtm, you can make the small change then merge

GuillaumeLagrange:
Don't forget to change the valgrind-codspeed branch once the PR is merged

PR #526, opened 2026-09-03
refactor(memtrack): rename track-rmap toggle to track-physical

not-matthias:
## What

Renames the memtrack physical-memory tracking toggle so it names the capability rather than the underlying mechanism:

- Env var: `CODSPEED_MEMTRACK_TRACK_RMAP` -> `CODSPEED_MEMTRACK_TRACK_PHYSICAL`
- Field: `TrackerOptions.rmap` -> `TrackerOptions.physical`

Physical (resident) memory is reconstructed from the folio rmap fentry hooks, so `physical` describes the intent while the internal rmap mechanism layer (`RmapSupport`, `MemtrackBpf::new_with_rmap`, `for_each_rmap_*`, the BPF C) keeps its accurate name.

## Behavior

**Unchanged.** Tracking stays opt-in (enable with `=1`); the builder default remains `false`. The full folio rmap hook set only exists on newer kernels, and `RmapSupport::detect()` still gates what actually attaches per kernel — so the default is not flipped to on.

## Notes

- Only the env-var/field names change; callers and the `rmap_only_options()` test helper are updated accordingly.
- `cargo check --tests -p memtrack` passes (crate + test targets compile). The privileged rss/c integration tests are `GITHUB_ACTIONS`-gated and skipped locally, so they were not executed here.

Closes COD-3468

### DIFF
--- a/crates/memtrack/src/ebpf/tracker.rs
+++ b/crates/memtrack/src/ebpf/tracker.rs
@@ -20,6 +20,9 @@ pub struct TrackerOptions {
     /// folio rmap hooks, which only attach on kernels that expose them.
     #[builder(default = false)]
     pub physical: bool,
+    /// Uprobe attach mechanism. `None` detects it from BPF token availability.
+    #[builder(default, setter(strip_option))]
+    pub variant: Option<BpfVariant>,
 }
 
 impl TrackerOptions {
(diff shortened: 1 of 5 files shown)


crates/memtrack/src/ebpf/tracker.rs
```diff
@@ -16,9 +16,13 @@ pub struct TrackerOptions {
     /// exec-mapping watcher.
     #[builder(default = true)]
     pub allocators: bool,
-    /// Reconstruct per-process RSS from the folio rmap fentry hooks.
+    /// Track physical (resident) memory: the `rss_stat` tracepoint for the
```
GuillaumeLagrange: I think we can make this comment even longer 🫠

GuillaumeLagrange (approved):
olgtm

crates/memtrack/src/ebpf/tracker.rs
not-matthias: Sorry about this one :grimacing:

GuillaumeLagrange: You are forgiven

PR #531, opened 2026-09-07
feat!: ship one binary — bundle exec-harness and memtrack, drop the LD_PRELOAD hack

moha-bekh:
`exec-harness` and `memtrack` are compiled into `codspeed` and reached as hidden subcommands, and
`exec-harness` no longer injects `libcodspeed_preload.so`. The two are one change: a statically
linked musl binary cannot be preloaded into a glibc process, so the single binary was blocked on
removing the preload.

One release artifact instead of three, 11.0 MB compressed against 13.5 MB today. The download
machinery and both installer pins are gone.

Three things worth a reviewer's attention:

- **The measurement baseline shifts.** The measured region now starts in exec-harness before the
  fork, so it also covers fork/exec/wait and the child's pre-`main` startup. It is a constant
  ~940k Ir per exec-harness benchmark, not a percentage — stable to 0.22% across a 1000× range of
  benchmark size. Shipping as is: no forced baseline, no history surgery.
- **Attribution rests entirely on the spawn-chain walk.** Children no longer self-label with the
  benchmark URI, since `LD_PRELOAD` is what used to be inherited by every descendant.
- **`setcap` now lands on `codspeed` itself**, so memtrack's five capabilities sit on the CLI.
  They are `+ep` with no inheritable set, so a spawned benchmark does not receive them.

Also removes the user-facing *"CPU Simulation mode does not support statically linked binaries"*
error: nothing is injected into the benchmarked executable any more.

Closes COD-3218
Closes COD-3440


crates/exec-harness/build.rs
```diff
@@ -1,170 +1,18 @@
...
-}
+/// Integration name reported to CodSpeed.
+const INTEGRATION_NAME: &str = "exec-harness";
 
 fn main() {
-    println!("cargo:rerun-if-changed=preload/codspeed_preload.c");
-    println!("cargo:rerun-if-env-changed=CODSPEED_INSTRUMENT_HOOKS_DIR");
-
-    let preload_constants: PreloadConstants = PreloadConstants::default();
+    println!("cargo:rerun-if-changed=build.rs");
```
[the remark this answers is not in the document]
moha-bekh: Not applicable here: `build.rs` is a build script, and `println!` to stdout is the only channel cargo reads. `cargo:rerun-if-changed=` and `cargo:rustc-env=` are directives interpreted by cargo, not log output — routing them through `tracing` would silently drop them and break the build (`CODSPEED_INTEGRATION_NAME` and `CODSPEED_INTEGRATION_VERSION` would stop being set, and `src/constants.rs` reads them with `env!()`).

The no-`println!` rule is about the CLI, where output must go through the logger. Worth scoping it to exclude `build.rs` files.

src/executor/memory/setup.rs
```diff
@@ -37,8 +35,14 @@ fn memtrack_setcap_spec() -> String {
...
 
+/// The binary that must carry the eBPF capabilities: memtrack runs as a
+/// subcommand of this executable, so it is this one.
+///
+/// Every invocation of the CLI therefore carries [`MEMTRACK_REQUIRED_CAPS`],
+/// `CAP_SYS_ADMIN` included. They are granted `+ep` and not inheritable, so a
+/// spawned benchmark does not receive them: the elevation stops here.
 fn memtrack_path() -> Option<PathBuf> {
-    which::which(MEMTRACK_COMMAND).ok()
+    self_exe().ok()
```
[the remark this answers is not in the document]
moha-bekh: Valid, and worse than described: `CODSPEED_SELF_EXE` had **no production caller at all**. It exists so the tests can point at the real binary, because under `cargo test` `current_exe()` is the test harness. The doc comment justified it with a launcher scenario that does not exist in this repo.

Fixed in 9cbe2e4 by putting the override behind `cfg(test)`, the constant included, so a release build always resolves `current_exe()` and there is nothing left to influence. The 65 executor tests still pass, which is what confirms nothing outside the tests relied on it.

src/cli/mod.rs
```diff
@@ -211,6 +256,9 @@ pub async fn run() -> Result<()> {
         Commands::Show => show::run()?,
         Commands::Update => update::run().await?,
         Commands::Internal(InternalCommands::Samply(args)) => samply::run(args)?,
+        Commands::Internal(InternalCommands::ExecHarness(args)) => exec_harness::run(args)?,
+        #[cfg(target_os = "linux")]
+        Commands::Internal(InternalCommands::Memtrack(args)) => memtrack::run(args)?,
```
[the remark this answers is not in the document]
moha-bekh: Valid. The re-exec runs in the **benchmark's** working directory, i.e. the user's project, so `DiscoveredProjectConfig::discover_and_load` could abort a measurement over a `codspeed.yaml` that has nothing to do with it.

Fixed in 90ee59c: the bundled subcommands are dispatched in `run_internal`, called right after `Cli::parse()`, before the profile config, the API client, the project discovery and the logger. They share none of that with the runner.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,9 @@ jobs:
       - uses: ./.github/actions/install-rust
         with:
           components: rustfmt, clippy
+      # Building the runner builds memtrack's vendored libbpf-sys with it.
+      - uses: ./.github/actions/install-bpf-deps
+        if: matrix.os == 'ubuntu-latest'
```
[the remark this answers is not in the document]
moha-bekh: Valid, and it is a gap this PR itself opened: a throwaway spike workflow covered the musl builds and was deleted with the spike.

Fixed in 9dd1b1e, which adds a permanent `musl-build` job to `ci.yml`: both distribution targets on native runners, `readelf`-based static assertions rather than a `file` string (rustc emits a static-PIE for x86_64 musl and spells it differently from aarch64), and `codspeed exec-harness --version` / `codspeed memtrack --version`, which only answer if both CLIs are really linked in. It builds with no environment variables, which doubles as the check that `.cargo/config.toml` carries the whole recipe. Both legs green on the first run.

GuillaumeLagrange (commented):
@not-matthias can you do a first round of review 🙏 ? Let's ignore the `LD_PRELOAD` removal overhead for the exec-harness simulation, I'll spec this in a dedicated ticket and we'll do it before we merge and release this.

crates/exec-harness/Cargo.toml
```diff
@@ -20,10 +20,7 @@ serde = { workspace = true }
...
 runner-shared = { path = "../runner-shared" }
 tempfile = { workspace = true }
-object = { workspace = true }
 
-[build-dependencies]
-cc = "1"
-
-[package.metadata.dist]
-targets = ["aarch64-unknown-linux-gnu", "x86_64-unknown-linux-gnu"]
+# Deliberately no `[package.metadata.dist]`: exec-harness ships inside the
```
not-matthias: Remove this comment, we don't need to document the non-existence of something ^^<br><br>(also, this would be a good test/use-case for the deslop skill)

crates/memtrack/Cargo.toml
```diff
@@ -50,17 +50,6 @@ test-log = { workspace = true }
...
-zlib1g-dev = "*"
-libbpf-dev = "*"
-
-# Required for the vendored feature
-autopoint = "*"
-bison = "*"
-flex = "*"
+# Deliberately no `[package.metadata.dist]`: memtrack ships inside the
+# `codspeed` binary and is not released on its own. Its apt build dependencies
+# belong to the root package, which is what cargo-dist builds.
```
not-matthias: same here, remove

CONTRIBUTING.md
```diff
@@ -10,50 +10,28 @@ prek install
...
-
-After releasing `memtrack` or `exec-harness`, you **must** update the version references in the runner code:
-
-1. **For memtrack**: Update the `MEMTRACK_INSTALLER` pin record in `src/binary_pins.rs` (see [Pinned binary hashes](#pinned-binary-hashes) below).
-
-2. **For exec-harness**: Update the `EXEC_HARNESS_INSTALLER` pin record in `src/binary_pins.rs`.
-
-These constants are used by the runner to download and install the correct versions of the binaries from GitHub releases.
+Both still keep their own `version` in `Cargo.toml` — that is what
+`codspeed exec-harness --version` reports — but bumping it is a plain edit, not a release.
```
not-matthias: deslop + simplify, way too verbose

crates/exec-harness/src/analysis/mod.rs
```diff
@@ -1,25 +1,60 @@
...
-use crate::constants;
 use crate::uri;
 use instrument_hooks_bindings::InstrumentHooks;
 use std::process::Command;
 
-mod ld_preload_check;
-mod preload_lib_file;
-
-pub fn perform(commands: Vec<BenchmarkCommand>) -> Result<()> {
+/// Executes the given benchmark commands, measuring each one through the
```
not-matthias: deslop

crates/exec-harness/src/analysis/mod.rs
```diff
@@ -1,25 +1,60 @@
...
+/// [`InstrumentHooks::set_executed_benchmark`] then names with the benchmark
+/// URI. The backend walks that edge to attribute the child's trace to the
+/// benchmark, so the measurement covers the whole spawned process tree.
+///
+/// Nothing is injected into the benchmarked executable, so statically linked
+/// ones work.
+pub fn perform(commands: Vec<BenchmarkCommand>, mode: MeasurementMode) -> Result<()> {
     let hooks = InstrumentHooks::instance(INTEGRATION_NAME, INTEGRATION_VERSION);
 
+    if !hooks.is_instrumented() {
```
not-matthias: are you sure we can actually hit this case? this seems like a case that shouldn't be reachable in the first place. i'd rather we fix the root cause in the runner than the symptom here.

crates/exec-harness/src/analysis/mod.rs
```diff
@@ -1,25 +1,60 @@
...
         let mut cmd = Command::new(&benchmark_cmd.command[0]);
         cmd.args(&benchmark_cmd.command[1..]);
+
+        if mode == MeasurementMode::Simulation {
+            // Make sure python and node processes output perf maps, so the
+            // runner can resolve JIT-ed frames afterwards. For python this is
+            // usually done by `pytest-codspeed`.
+            cmd.env("PYTHONPERFSUPPORT", "1");
+            crate::node::set_node_options(&mut cmd);
+        }
```
not-matthias, line 30: Shouldn't this be also set in Walltime?

crates/instrument-hooks-bindings/build.rs
```diff
@@ -35,6 +37,24 @@ fn main() {
...
+                    "Failed to compile the instrument-hooks native library with cc-rs.\n\
+                     A Linux build must not fall back to the noop implementation: it \
+                     would run benchmarks and measure nothing.\n\
+                     Make sure a C compiler for the target is installed and reachable \
+                     by cc-rs (for musl targets, `musl-tools` provides \
+                     `<arch>-linux-musl-gcc`).\n\
+                     Compiler information: {compiler:?}\n\
+                     Compilation error: {e}"
+                );
+            }
```
not-matthias, line 56: We also measure on MacOS, and soon on Windows. I'd prefer if we don't try to modify the behavior of the exec-harness in this PR as it's already big enough.

crates/exec-harness/src/cli.rs
```diff
@@ -0,0 +1,65 @@
+//! exec-harness's command line, shared by the standalone `exec-harness` binary
+//! and by the `codspeed` CLI that bundles it as a hidden subcommand.
+//!
+//! Keeping the parser and the dispatch here rather than in `main.rs` is what
+//! stops the two paths from drifting: the standalone binary is a wrapper over
+//! [`run_cli`] and nothing else. It deliberately does **not** install a logger —
+//! only one global logger can exist per process, and when exec-harness runs
+//! bundled the host CLI has already installed its own.
```
not-matthias: This comment shouldn't be here. Pls deslop. We shouldn't document this type of stuff

crates/exec-harness/src/main.rs
```diff
@@ -1,33 +1,12 @@
-use clap::Parser;
+//! The standalone `exec-harness` binary.
```
not-matthias: deslop/remove<br><br>we should NEVER have comments like this in the code:

> ```
> //! All that is left here is the global logger, which only makes sense when
> //! exec-harness owns the whole process.
> ```

crates/memtrack/src/lib.rs
```diff
@@ -1,5 +1,10 @@
 mod allocators;
 mod bpf_token;
+/// The command line, shared by the standalone binary and the bundled subcommand.
```
not-matthias: remove

src/cli/mod.rs
```diff
@@ -110,37 +113,96 @@ pub(crate) enum InternalCommands {
...
+#[cfg(test)]
 pub(crate) const SELF_EXE_ENV_VAR: &str = "CODSPEED_SELF_EXE";
 
+/// The executable that internal subcommands are re-invoked through.
+///
+/// Exposed separately from [`InternalCommands::get_command_builder`] because the
+/// memory executor grants file capabilities to this exact path before running
+/// it, and `setcap` on a path that is not the one later exec'd succeeds while
+/// changing nothing.
+pub(crate) fn self_exe() -> Result<PathBuf> {
```
not-matthias, line 136: why was this made a func and not left in get_Command_builder?

src/cli/exec/multi_targets.rs
```diff
@@ -92,9 +94,43 @@ pub fn build_exec_targets_pipe_command(
...
+    Ok(build_pipe_command_from_json(exec_harness, &json))
+}
+
+fn build_pipe_command_from_json(exec_harness: &str, json: &str) -> String {
+    format!("{exec_harness} - <<'CODSPEED_EOF'\n{json}\nCODSPEED_EOF")
 }
 
-fn build_pipe_command_from_json(json: &str) -> String {
-    format!("{EXEC_HARNESS_COMMAND} - <<'CODSPEED_EOF'\n{json}\nCODSPEED_EOF")
+#[cfg(test)]
```
not-matthias, line 104: i think we don't really need those tests. we already have integration tests that validate that it works.

src/executor/helpers/command.rs
```diff
@@ -128,7 +128,12 @@ impl CommandBuilder {
         self
     }
 
-    /// Returns the command line as a string for debugging/testing purposes
+    /// The command line as a single shell-quoted string.
+    ///
+    /// Used for logging and assertions, and by
+    /// [`InternalCommands::get_shell_command`](crate::cli::InternalCommands::get_shell_command)
+    /// to splice a re-exec into a command that `bash -c` runs — so the quoting
+    /// has to stay correct, not merely readable.
```
not-matthias: why was this comment changed even though the func wasn't touched?

src/executor/config.rs
```diff
@@ -193,12 +193,23 @@ impl OrchestratorConfig {
...
+    ///   state across `fork`/`exec`, which `--instr-atstart=inherit` enables;
+    ///   with `no` it dumps a single zero-cost part and reports nothing.
+    ///
+    /// Entrypoint runs must keep the opposite default, or a benchmark that forks
+    /// would silently start counting its children.
     pub fn executor_config_for_command(
         &self,
         command: String,
-        enable_introspection: bool,
+        uses_exec_harness: bool,
```
not-matthias, line 205: This is a breaking change! ⚠️⚡<br><br>`simulation_track_subprocess` is currently off-by-default and enabling it for exec harness could cause CI to take longer and might change benchmark results. Try to keep the same logic as we had it before.

crates/memtrack/src/ebpf/memtrack/mod.rs
```diff
@@ -282,19 +282,62 @@ impl Drop for MemtrackBpf {
...
     fn libc_allocator_symbols_resolve_to_offsets() {
-        let maps = std::fs::read_to_string("/proc/self/maps").unwrap();
-        let libc_path = maps
-            .lines()
-            .find_map(|line| {
-                let path = line.split_whitespace().last()?;
-                path.contains("libc.so.6").then(|| path.to_owned())
-            })
-            .expect("test process has no mapped libc.so.6");
+        let libc_path = libc_mapped_by_a_child();
```
not-matthias: That's a lot of code for just getting the path to a libc file. let's either:

1. use `ldd` instead
2. or hardcode the path (since it's the same on ubuntu) and only run this test in CI and not locally using test_with

src/executor/tests.rs
```diff
@@ -175,6 +175,29 @@ fi
         (execution_context, temp_dir)
     }
 
+    /// Serializes every executor test that drives a benchmark through the runner
```
not-matthias: This change isn't really needed, no? In that case it shouldn't be part of this PR. Please correct me if I'm wrong, didn't check that closely

src/executor/memory/setup.rs
```diff
@@ -94,73 +98,19 @@ pub fn ensure_memtrack_capabilities() -> Result<()> {
...
+        tool_name: MEMTRACK_COMMAND.to_string(),
+        status: ToolInstallStatus::Installed {
+            version: env!("CARGO_PKG_VERSION").to_string(),
         },
     }
 }
 
+/// No-op: memtrack is part of this binary. Kept so the setup flow can treat it
+/// like the tools that do need installing.
 pub async fn install_memtrack() -> Result<()> {
```
not-matthias: no need to log anything, the func can be removed

src/run_environment/local/provider.rs
not-matthias, line 1: why was this changed?

not-matthias (changes requested):
Overall pretty good, just a few comments on how to better structure the code/comments

And a few notes on how to best structure the PRs for reviewers:

* Try to keep the changes as minimal as possible
* Do not modify unrelated comments/code -> should be done in separate PRs for easier review
* Remove obvious/LLM-written/bloated comments (using the deslop skill)
* you should also review the PR before putting it into review

crates/exec-harness/src/analysis/mod.rs
moha-bekh: You're right, it isn't reachable through the runner. The root cause was on our side: InternalCommands had no hide = true, so codspeed --help advertised exec-harness and memtrack as regular commands. Hidden them (samply too) and dropped the check.

crates/memtrack/src/ebpf/memtrack/mod.rs
moha-bekh: Done — hardcoded the Ubuntu path and gated it on GITHUB_ACTIONS, 51 lines gone. Built from std::env::consts::ARCH since bpf-tests also runs on the arm runners.

src/executor/tests.rs
moha-bekh: Split it. The shared FIFO lock was only a test-level workaround for a pre-existing race (COD-3560) — moving it to its own PR, walltime is now identical to main.

The rest is caused by this PR: it drops the cargo install memtrack and setup --mode memory CI steps, so the tests have to setcap themselves — through SELF_EXE_ENV_VAR, since current_exe() under cargo test is the harness. The sudo guard comes with it: capabilities are an xattr, so the grant is lost on every relink, and a hidden sudo prompt under cargo test hangs forever.

src/executor/config.rs
moha-bekh: Split the two flags. --separate-threads stays on simulation_track_subprocess alone, so nothing changes there for anyone.

--instr-atstart=inherit I can't revert: without LD_PRELOAD the child no longer instruments itself, exec-harness toggles in its own process and forks, so no makes the run dump a single zero-cost part and report nothing. It also isn't "track subprocesses" in the user-facing sense here — the child is the benchmark. Entrypoint runs keep the old default untouched.

src/run_environment/local/provider.rs
moha-bekh: No good reason — it's a test helper that turns .output().unwrap() into a real assertion on git's exit status, so a broken setup doesn't resurface later as an UnbornBranch error. Useful, but unrelated to this PR. Reverted, I'll send it separately.

crates/exec-harness/src/analysis/mod.rs
```diff
@@ -1,17 +1,19 @@
...
 
-mod ld_preload_check;
-mod preload_lib_file;
-
-pub fn perform(commands: Vec<BenchmarkCommand>) -> Result<()> {
+/// Runs each benchmark command with the instrumentation toggled around its spawn.
+///
+/// The toggles are in *this* process: the child inherits the state across
+/// `fork`/`exec`, and callgrind records the spawn edge on the dump part that
+/// [`InstrumentHooks::set_executed_benchmark`] names with the benchmark URI.
```
GuillaumeLagrange: Slop comment

src/executor/valgrind/measure.rs
```diff
@@ -33,11 +33,18 @@ fn get_valgrind_args(tool: &SimulationTool, config: &ExecutorConfig) -> Vec<Stri
     .map(|x| x.to_string())
     .collect();
 
-    if config.simulation_track_subprocess {
+    // exec-harness toggles the instrumentation in its own process and then forks
+    // the benchmark, so the child is measured only if valgrind propagates that
+    // state across `fork`/`exec`. With `no` the run dumps a single zero-cost part.
+    if config.simulation_track_subprocess || config.uses_exec_harness {
```
GuillaumeLagrange: I think this should be resolved earlier in `executor_config_for_command`, and set the value directly in `simulation_track_subprocess`, and then we can remove the `uses_exec_harness` from the executor config, cause ideally it should be transparent for the executor

src/executor/valgrind/measure.rs
```diff
@@ -33,11 +33,18 @@ fn get_valgrind_args(tool: &SimulationTool, config: &ExecutorConfig) -> Vec<Stri
...
-        args.push("--separate-threads=yes".to_string());
     } else {
         args.push("--instr-atstart=no".to_string());
+    }
+
+    if config.simulation_track_subprocess {
+        args.push("--separate-threads=yes".to_string());
+    } else {
         args.push("--separate-threads=no".to_string());
     }
```
GuillaumeLagrange, line 42: Let's keep everything bundled together to avoid making the logic too complex for something that will become default behavior anyway next major (the subprocess tracking + thread costs)

src/executor/valgrind/measure.rs
```diff
@@ -33,11 +33,18 @@ fn get_valgrind_args(tool: &SimulationTool, config: &ExecutorConfig) -> Vec<Stri
     .map(|x| x.to_string())
     .collect();
 
-    if config.simulation_track_subprocess {
+    // exec-harness toggles the instrumentation in its own process and then forks
+    // the benchmark, so the child is measured only if valgrind propagates that
+    // state across `fork`/`exec`. With `no` the run dumps a single zero-cost part.
```
GuillaumeLagrange: This comment could be shortened, and put in `executor_config_for_command` where we force `simulation_track_subprocess` to true

crates/exec-harness/build.rs
```diff
@@ -1,170 +1,11 @@
...
-    /// Environment variable name for the benchmark URI.
-    uri_env: &'static str,
-    /// Integration name reported to CodSpeed.
-    integration_name: &'static str,
-    /// Integration version reported to CodSpeed.
-    integration_version: &'static str,
-    /// Filename for the preload shared library.
-    preload_lib_filename: &'static str,
-}
+const INTEGRATION_NAME: &str = "exec-harness";
```
GuillaumeLagrange: This no longer need to be set here as well since there is no shared library, we can put this as a const inside the code

crates/exec-harness/src/constants.rs
```diff
@@ -1,12 +1,3 @@
...
-//!
-//! These constants are defined in the build script (build.rs) and exported as
-//! environment variables. The same values are passed to the C preload library
-//! as compiler defines, ensuring both Rust and C code use the same source of truth.
-
-/// Environment variable name for the benchmark URI.
-pub const URI_ENV: &str = env!("CODSPEED_URI_ENV");
-
 /// Integration name reported to CodSpeed.
 pub const INTEGRATION_NAME: &str = env!("CODSPEED_INTEGRATION_NAME");
```
GuillaumeLagrange: See comment in build.rs: no point in using a build time env var here

src/cli/memtrack.rs
```diff
@@ -0,0 +1,19 @@
...
+}
+
+pub fn run(args: MemtrackArgs) -> Result<()> {
+    // memtrack's own clap parser expects its name as `argv[0]`, not ours.
+    let argv = std::iter::once(std::ffi::OsString::from("memtrack")).chain(args.args);
+
+    // memtrack's exit code is the tracked command's own, and the runner reads
+    // it to decide whether the benchmark failed, so it has to become ours.
+    let code = ::memtrack::cli::run_cli(argv)?;
+    std::process::exit(code);
```
GuillaumeLagrange: A bit weird that we have different behaviors between this and exec-harness regarding the exit code, any way we could align them?

src/cli/mod.rs
```diff
@@ -158,7 +208,8 @@ pub async fn run() -> Result<()> {
     let setup_cache_dir = setup_cache_dir.as_deref();
 
     match cli.command {
-        Commands::Run(_) | Commands::Exec(_) | Commands::Internal(InternalCommands::Samply(_)) => {} // these are responsible for their own logger initialization
+        // These initialize their own logging.
+        Commands::Run(_) | Commands::Exec(_) => {}
```
GuillaumeLagrange, line 210: Why did samply get removed from being responsible for its own logging? Just asking to make sure it's not a forgotten side effect

src/executor/config.rs
```diff
@@ -131,6 +131,9 @@ pub struct ExecutorConfig {
     /// Whether to enable language-level introspection (Node.js, Go wrappers in PATH).
     /// Disabled for exec-harness targets since they don't need it.
     pub enable_introspection: bool,
+    /// Whether this execution is driven by exec-harness rather than by a plain
+    /// entrypoint command.
+    pub uses_exec_harness: bool,
```
GuillaumeLagrange: As told in other comments, I do not think that's necessary, we should just override the `simulation_track_subprocess` where relevant

.cargo/config.toml
```diff
@@ -0,0 +1,24 @@
...
+ac_cv_search_fts_close = "none required"
+
+# The `argp.h` stub. `CPATH` and not `CFLAGS -I<path>`, because `relative = true`
+# only works on a bare path. The stub defers to the real header under glibc.
+CPATH = { value = "crates/memtrack/musl", relative = true }
+
+# Debian's musl-gcc runs with -nostdinc, so the kernel UAPI headers libbpf needs
+# have to be added back, last so a glibc build is unaffected. Both triplets are
+# listed because `[env]` cannot branch on the arch; a missing one is ignored.
+CFLAGS = "-idirafter /usr/include/x86_64-linux-gnu -idirafter /usr/include/aarch64-linux-gnu -idirafter /usr/include"
```
GuillaumeLagrange: Is there a way to set these ONLY for the library that needs it? I'd like us to avoid side effects.

It could also be quite helpful to link to github issues that led us to these solutions/why we need to do this in the first place!

.github/workflows/ci.yml
```diff
@@ -153,6 +146,50 @@ jobs:
           mode: ${{ matrix.mode }}
           run: cargo codspeed run -p runner-shared
 
+  # The released Linux artifacts are musl, and nothing else here builds them.
```
GuillaumeLagrange: I think we could remove this comment

GuillaumeLagrange (changes requested):
See comments for details.

I also see that there are unresolved comments.

When you handle a review, all comments should either be

* resolved and handled properly (very important to not silently resolve comments)
* answered with either counter arguments or questions so the reviewer can either accept your arguments or open a discussion

Finally, there are 35 commits in the PR, you may want to re-do commits before the next round of reviews

.github/workflows/ci.yml
GuillaumeLagrange: I think we could remove this comment

GuillaumeLagrange (changes requested):
See comments for details.

I also see that there are unresolved comments.

When you handle a review, all comments should either be

* resolved and handled properly (very important to not silently resolve comments)
* answered with either counter arguments or questions so the reviewer can either accept your arguments or open a discussion

Finally, there are 35 commits in the PR, you may want to re-do commits before the next round of reviews

moha-bekh:
Yeah I didn't see them since I needed to click on the see more button, I will check If they are already resolved, and try to reduce the amount of commits by squashing where it's coherent

crates/exec-harness/src/cli.rs
moha-bekh: the module doc is gone.

crates/exec-harness/src/main.rs
moha-bekh: also removed.

crates/memtrack/src/lib.rs
moha-bekh: same

src/executor/helpers/command.rs
moha-bekh: as_command_line isn't touched by this branch, so its doc is back to what it is on main. Same pass restored the valgrind and executor docs this branch had rewritten for no reason.

src/cli/exec/multi_targets.rs
moha-bekh: removed.

src/cli/mod.rs
moha-bekh: Because it has a second caller: `memtrack_path()` in
`src/executor/memory/setup.rs`. The memory executor `setcap`s that exact path
before the re-exec, and `setcap` on a path that isn't the one later exec'd
succeeds while changing nothing — so both sites have to resolve the executable
the same way. Inlining it in `get_command_builder` would mean duplicating the
`current_exe()` resolution in the memory setup.

crates/instrument-hooks-bindings/build.rs
moha-bekh: For the record, since it'll come back in its own PR: the `panic!` was gated on
`CARGO_CFG_TARGET_OS == "linux"`, so macOS (and Windows later) kept the warning
and the noop fallback. What it caught is a musl build with no `musl-gcc` on
PATH, which compiled the noop hooks and then ran benchmarks measuring nothing at
exit code 0.

crates/exec-harness/src/analysis/mod.rs
moha-bekh: `NODE_OPTIONS` already is: walltime goes through `set_node_options` in
`walltime/benchmark_loop.rs`. `PYTHONPERFSUPPORT` isn't, and wasn't on main
either — only `perform_with_valgrind` set it, and this branch just carries that
over to the simulation branch of `perform`. So it's a pre-existing gap in
walltime rather than something this PR changes; I'd rather not touch walltime
here. If you want I can open a ticket for it.

crates/exec-harness/Cargo.toml
moha-bekh: removed

crates/memtrack/Cargo.toml
moha-bekh: gone

CONTRIBUTING.md
moha-bekh: desloped

crates/exec-harness/src/analysis/mod.rs
moha-bekh: done

src/executor/memory/setup.rs
moha-bekh: done

crates/instrument-hooks-bindings/build.rs
moha-bekh: Taken out: f539a42 is reverted, the file is byte-identical to main again.

For the record, since it'll come back in its own PR: the `panic!` was gated on
`CARGO_CFG_TARGET_OS == "linux"`, so macOS (and Windows later) kept the warning
and the noop fallback. What it caught is a musl build with no `musl-gcc` on
PATH, which compiled the noop hooks and then ran benchmarks measuring nothing at
exit code 0.

crates/exec-harness/src/constants.rs
moha-bekh: Done, see the reply on build.rs.

crates/exec-harness/build.rs
moha-bekh: Right — the build script is deleted. `INTEGRATION_NAME` is a plain literal now
and `INTEGRATION_VERSION` is `env!("CARGO_PKG_VERSION")`, both in `constants.rs`.

crates/exec-harness/src/analysis/mod.rs
moha-bekh: Gone.

src/cli/memtrack.rs
moha-bekh: The difference is between the two tools, not between the two wrappers:
`memtrack::cli::run_cli` returns `Result<i32>` — the tracked command's own exit
code — while `exec_harness::cli::run_cli` returns `Result<()>` and bails when the
benchmark exits non-zero. Each bundled subcommand does what its standalone binary
already did on main: `memtrack`'s `main()` ended on `process::exit(code)`,
exec-harness's returned the `Result`.

Aligning them means changing one of the two contracts with the runner: either
exec-harness starts forwarding the child's exit code instead of erroring, or
memtrack stops forwarding it and the code is lost. Both are behavior changes I'd
rather not smuggle into this PR — happy to do it in a follow-up if you have a
preference on which way.

src/cli/mod.rs
moha-bekh: Not a side effect. Internal commands now return early:

    if let Commands::Internal(command) = cli.command {
        return run_internal(command);
    }

which happens before this match, so samply never reaches `init_local_logger()` —
exactly like before, when the `Internal(Samply(_))` arm was what skipped it. The
arm now only lists `Run` and `Exec` because those are the only non-internal
commands that own their logger.

.github/workflows/ci.yml
moha-bekh: Removed.

moha-bekh: Removed.

src/executor/valgrind/measure.rs
moha-bekh: I'll do the refactor from the other three comments. The bundling I'd like
confirmed first: it undoes f1110dd, which @not-matthias asked for when he flagged
forcing `simulation_track_subprocess` on as a breaking change (off by default
today, slower CI, results can move).

That's why the flags are split: exec-harness forces `--instr-atstart=inherit`
only, which it can't run without, and `--separate-threads` stays where it was.
Bundling them back turns it on for exec-harness too. Your call about the next
major is a fair argument — settle it with him and I'll land all four in one
commit.

.cargo/config.toml
moha-bekh: Done for the flags: `CFLAGS` is now `CFLAGS_<target>` for both musl triples.
cc-rs reads the target-scoped name first and libbpf-sys builds zlib, elfutils and
libbpf from `cc::Tool::cflags_env()`, so a gnu build never sees it.

`CPATH` has to stay global: the stub is for elfutils, which re-derives its flags
from `cflags_env()`, and a `-I` there would need an absolute path (`relative =
true` only works on a bare value). The stub defers to the real `<argp.h>` under
`__GLIBC__`. The `ac_cv_*` names are only read by autoconf.

Upstream issue linked in the file: libbpf/libbpf-sys#153

src/executor/valgrind/measure.rs
moha-bekh: On the threads I left open on purpose: four are answered and waiting on you two
rather than on me — the walltime `PYTHONPERFSUPPORT`, `self_exe()`, the exit
codes, and the `--separate-threads` bundling. The three on `measure.rs` and
`executor/config.rs` are the same change, and it hangs on that last one: I'll
land them in one commit once you've settled it.

src/executor/valgrind/measure.rs
not-matthias: As discussed, in order to remove LD_PRELOAD, let's enable simulation_track_subprocess when exec harness is used.

crates/exec-harness/src/analysis/mod.rs
not-matthias: ALright, let's keep it like this for now

src/executor/memory/executor.rs
```diff
@@ -146,7 +147,8 @@ impl Executor for MemoryExecutor {
         _system_info: &SystemInfo,
         _setup_cache_dir: Option<&Path>,
     ) -> Result<()> {
-        install_memtrack().await
+        // memtrack ships inside this binary, nothing to install.
+        Ok(())
```
not-matthias: no need to even impl this, we can remove it entirely and use the default trait impl

src/executor/memory/setup.rs
```diff
@@ -1,15 +1,12 @@
...
 use crate::executor::helpers::run_with_sudo::{is_root_user, run_with_sudo};
 use crate::executor::{ToolInstallStatus, ToolStatus};
 use crate::prelude::*;
 use caps::Capability;
 use std::path::PathBuf;
-use std::process::Command;
 
-pub const MEMTRACK_COMMAND: &str = "codspeed-memtrack";
-pub const MEMTRACK_CODSPEED_VERSION: &str = binary_pins::MEMTRACK_VERSION;
+pub const MEMTRACK_COMMAND: &str = "memtrack";
```
not-matthias, line 8: i think we can inline this const, as it's just used in a single place

src/executor/memory/setup.rs
```diff
@@ -94,73 +94,11 @@ pub fn ensure_memtrack_capabilities() -> Result<()> {
...
-
+    // memtrack ships inside this binary, so it is installed by construction.
     ToolStatus {
-        tool_name,
-        status: ToolInstallStatus::IncorrectVersion {
-            version,
-            message: "could not parse version".to_string(),
+        tool_name: MEMTRACK_COMMAND.to_string(),
+        status: ToolInstallStatus::Installed {
+            version: env!("CARGO_PKG_VERSION").to_string(),
```
not-matthias: I think we should still expose the `memtrack --version` , not the codspeed version. 
wdyt @GuillaumeLagrange?

not-matthias (commented):
we're getting there!

src/executor/memory/setup.rs
GuillaumeLagrange: I think this should go altogether, memtrack version has absolutely no interest here now that we have one binary

not-matthias (commented):
we're getting there!

src/executor/memory/setup.rs
moha-bekh: It's used in 5 places (4 logs in ensure_memtrack_capabilities + the bail! in ensure_privileges), so inlining would put the literal back in 5 format strings, You still want that I inline it ?

src/executor/memory/setup.rs
moha-bekh: Removed, tool_status() returns None — I also hoisted the privilege_status() block out of the Installed branch in status(), otherwise the capability state would have disappeared from setup --mode memory status.

src/executor/memory/executor.rs
moha-bekh: Removed.

src/executor/valgrind/measure.rs
moha-bekh: Done — executor_config_for_command now forces simulation_track_subprocess when the target uses exec-harness, ExecutorConfig no longer carries uses_exec_harness. get_valgrind_args is back to a single branch, so measure.rs is now identical to main and exec-harness runs get --separate-threads=yes as well.

src/executor/valgrind/measure.rs
moha-bekh: Same change, answered in https://github.com/CodSpeedHQ/codspeed/pull/531#discussion_r4070533450

src/executor/valgrind/measure.rs
moha-bekh: Same change -> https://github.com/CodSpeedHQ/codspeed/pull/531#discussion_r4070533450

src/executor/config.rs
moha-bekh: Same as https://github.com/CodSpeedHQ/codspeed/pull/531#discussion_r4070533450

PR #536, opened 2026-09-16
fix(memtrack): attach lifecycle and syscall hooks without perf_event_open

GuillaumeLagrange:


crates/memtrack/src/ebpf/c/allocator.h
not-matthias, line 1: These syscalls are actually not yet used in the backend. We could also just remove them for now, until we actually decide to model them.

not-matthias, line 1: And so the main issue was that we didn't enable those tracepoints in earlier releases, right? In this case, we actually had a lot of dead code in eBPF that could've been removed. 

Might be useful to setup a Greptile Rule (https://www.greptile.com/docs/code-review/greptile-config) in memtrack to try and detect this. E.g. something along the lines of: 
```bash
# .greptile/rules.md
NEVER allow dead code in eBPF unless explicitly allowed by the user. Tracepoints/Uprobes/etc. that aren't used should be removed.
```

Lmk what you thknk

not-matthias (approved):
LGTM

crates/memtrack/src/ebpf/c/allocator.h
```diff
@@ -5,6 +5,16 @@
 #include "utils/map_helpers.h"
 #include "utils/process_tracking.h"
 
+/* Syscall entry points are arch-prefixed wrappers around the SYSCALL_DEFINE
+ * body (`__x64_sys_mmap`, `__arm64_sys_mmap`), and fentry targets a symbol. */
+#if defined(__TARGET_ARCH_x86)
+#define SYSCALL_SYM(name) "__x64_sys_" #name
+#elif defined(__TARGET_ARCH_arm64)
+#define SYSCALL_SYM(name) "__arm64_sys_" #name
```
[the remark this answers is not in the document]
GuillaumeLagrange: Dont care here

crates/memtrack/src/ebpf/c/allocator.h
GuillaumeLagrange: I'll remove them then!

GuillaumeLagrange: As discussed, I think it's a bit hard to enforce unless we also have greptile review bleed into the backend, which I'm not really sure we do.

I've removed the syscalls altogether, and we can evaluate how sueful such a rule would be if we ever lose track of bpf programs

PR #537, opened 2026-09-16
ci: add memtrack walltime benchmarks to CI

not-matthias:
Adds a `memtrack-benchmarks` CI job that runs `codspeed-memtrack track` against a few representative workloads (`ls`, `dd`, `tar`), each with RSS-only and RSS+rmap variants, to catch regressions in memtrack's own overhead.

Runs both `walltime` (execution time) and `memory` (memtrack's own peak RSS/allocations while tracking each workload) modes against the same `crates/memtrack/codspeed.yml` config.


.github/workflows/ci.yml
```diff
@@ -153,6 +153,49 @@ jobs:
...
+
+      - name: Prepare memtrack output directory
+        run: mkdir -p /tmp/codspeed-memtrack-bench
+
+      # The write benchmarks overwrite these archives every round; creating
+      # them inside a measured round would time one-off disk allocation.
+      - name: Pre-create benchmark archives
+        run: |
+          tar -cf /tmp/memtrack-bench.tar /usr/bin
+          dd if=/dev/zero of=/tmp/memtrack-bench-dd.bin bs=1M count=64 2> /dev/null
```
GuillaumeLagrange, line 212: Would be a cool usecase for "Before all" or something in the exec_harness 🤓

crates/memtrack/codspeed.yml
```diff
@@ -0,0 +1,43 @@
...
+# RSS-only and with-physical variants run back to back so physical
+# tracking's overhead is directly comparable.
+#
+# The warmup/max times are generous because a single tracked run already pays a
+# fixed BPF load + uprobe attach cost, which is far above the defaults tuned
+# for near-instant commands.
+#
+# The write workloads overwrite archives pre-created by CI before measurement:
+# a first-time multi-GB allocation on the runner disk would dominate one
+# variant's warmup round purely from I/O ordering, not tracking overhead.
```
GuillaumeLagrange: This is the only interesting paragraph here IMO, potentially the second one as well depending on how much of it is vibes

crates/runner-shared/benches/memtrack_writer.rs
```diff
@@ -139,3 +141,77 @@ fn encode_events_realistic(bencher: Bencher, n_workers: usize) {
         output
     });
 }
+
+/// Matches the batch size the ring-buffer poller flushes with, so the batched
+/// channel bench reproduces the production send pattern.
+const CHANNEL_BATCH: usize = 1024;
```
GuillaumeLagrange: I know this is bench code, but it's always a bit weird to introduce indirect coupling of two consts like this, high chances they somehow drift at one point

GuillaumeLagrange (approved):
olgtm, good idea!

GuillaumeLagrange:
Memtrack walltime benchmarks took 19m to run, maybe we can shard something before introducing a 20m delay in the CI pipeline? 😇

crates/runner-shared/benches/memtrack_writer.rs
not-matthias: Agreed, I decided to remove it since it's more about testing different variations of the code. It's more a side effect of optimization, rather than something we actually want. I initially thought it might be useful to keep, but let's just remove it.

.github/workflows/ci.yml
```diff
@@ -153,6 +153,72 @@ jobs:
...
+        run: |
+          mkdir -p /tmp/codspeed-memtrack-bench
+          tar -cf /tmp/memtrack-bench.tar /usr/bin
+          dd if=/dev/zero of=/tmp/memtrack-bench-dd.bin bs=1M count=64 2> /dev/null
+
+      - name: Run memtrack walltime benchmarks (${{ matrix.workload }})
+        uses: CodSpeedHQ/action@3194d9a39c4d46684cb44bf7207fc56626aad8fd # v4
+        with:
+          config: crates/memtrack/benchmarks/${{ matrix.workload }}.yml
+          mode: walltime
```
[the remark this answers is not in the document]
not-matthias: This is expected. Tracking memtrack within memtrack currently doesn't work reliably and is not something we need.

PR #539, opened 2026-09-16
fix: detect free-threaded Python from venv and uv before setting PYTHONMALLOC

not-matthias:
## Problem

Free-threaded CPython (3.13t+) refuses to start with `PYTHONMALLOC=malloc`. The valgrind executor already skips that override when it detects a free-threaded interpreter, but the detection only probed `python` on the `PATH`. It missed the interpreter the benchmark actually runs under when that interpreter lives in a virtual environment or is selected by `uv`, so runs using `astral-sh/setup-uv` with `python-version: 3.13t` still failed.

## Change

Probe every interpreter the benchmark command could resolve to, and skip the override if any of them is free-threaded:

- `python` / `python3` on the `PATH`
- `$VIRTUAL_ENV/bin/python`
- `.venv/bin/python` in the benchmark working directory
- the interpreter `uv python find $UV_PYTHON` resolves to (`setup-uv` exports `UV_PYTHON` from its `python-version` input). `uv` downloads interpreters lazily, so when the request is not installed yet the request string itself (`3.13t`, `+freethreaded`) decides.

The probe now reads `sys.abiflags` instead of `sysconfig.get_config_var('Py_GIL_DISABLED')`. `import sysconfig` fails when `_PYTHON_SYSCONFIGDATA_NAME` is set for a different interpreter (common in Nix shells), which turned the old check into a silent false negative for exactly the venvs it needs to detect.

## Verification

Release-mode smoke runs against real interpreters created with `uv venv`:

| scenario | free-threaded detected |
|---|---|
| no venv, PATH python 3.12 | no |
| plain 3.12 `.venv` in cwd | no |
| 3.14t `.venv` in cwd | yes |
| `UV_PYTHON=3.14t` (installed), no venv | yes |
| `UV_PYTHON=3.13t` (not installed), no venv | yes (request fallback) |
| `UV_PYTHON=3.12`, no venv | no |
| `VIRTUAL_ENV=<3.14t venv>`, cwd has no venv | yes |

Plus `rstest` cases for the `UV_PYTHON` request-string classifier.


not-matthias:
NOTE: This PR tries to fix the known cases when a free-threaded python can be used and try to not use the environment variable. However, there are many more cases that can bypass our checks (e.g. using custom compiled python). 

In the future, we can explore removing `PYTHONMALLOC`, but this would be a breaking change so it should only be included in the next runner release (as to why this partial solution exists)

src/executor/valgrind/helpers/python.rs
```diff
@@ -1,21 +1,103 @@
...
+/// `uv` downloads a requested interpreter lazily, so when `$UV_PYTHON` is not installed
+/// yet the request string itself decides.
+pub fn is_free_threaded_python(working_directory: Option<&Path>) -> bool {
+    let cwd = working_directory.unwrap_or(Path::new("."));
+
+    let mut candidates: Vec<PathBuf> = vec![PathBuf::from("python"), PathBuf::from("python3")];
+    if let Some(venv) = std::env::var_os("VIRTUAL_ENV") {
+        candidates.push(Path::new(&venv).join("bin/python"));
+    }
+    candidates.push(cwd.join(".venv/bin/python"));
```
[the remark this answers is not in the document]
GuillaumeLagrange: I think this is out of scope, we are doing best-effort stuff here, if people want to manually call python by absolute path so be it

src/executor/valgrind/helpers/python.rs
```diff
@@ -1,21 +1,103 @@
...
-
-    match output {
-        Ok(output) if output.status.success() => {
-            let stdout = String::from_utf8_lossy(&output.stdout);
-            stdout.trim() == "1"
+use crate::prelude::*;
+
+/// Free-threaded builds carry the `t` ABI flag. `sys.abiflags` needs no `sysconfig`
+/// import, which fails when `_PYTHON_SYSCONFIGDATA_NAME` names a module the
+/// interpreter does not ship.
```
GuillaumeLagrange: Why did we switch from a dynamic in-python detection that is recommended by the documentation? Did you run into explicit issue or is this LLM overengineering?

src/cli/experimental.rs
```diff
@@ -27,6 +27,15 @@ pub struct ExperimentalArgs {
...
     pub experimental_memory_track_physical: bool,
 
+    /// Do not set PYTHONMALLOC for simulation runs.
+    #[arg(
+        long,
+        default_value_t = false,
+        help_heading = "Experimental",
+        env = "CODSPEED_EXPERIMENTAL_DISABLE_PYTHONMALLOC"
+    )]
+    pub experimental_disable_pythonmalloc: bool,
```
GuillaumeLagrange: Not a fan of the `disable_pythonmalloc` flag name, wouldnt something like `do_not_force_python_malloc` be a bit more explicit?<br><br>Or `disable_pythonmalloc_override`

GuillaumeLagrange (approved):
olgtm

src/cli/experimental.rs
not-matthias: Changed to disable_pythonmalloc_override

src/executor/valgrind/helpers/python.rs
not-matthias: Good catch, I didn't see this. This seems to have been caused by an issue in my Nix flake which made the clanker change it. Obviously not what we want, so I reverted it back to the original

not-matthias:
@greptileai review

src/executor/valgrind/helpers/python.rs
```diff
@@ -1,21 +1,102 @@
...
-
-    match output {
-        Ok(output) if output.status.success() => {
-            let stdout = String::from_utf8_lossy(&output.stdout);
-            stdout.trim() == "1"
+use crate::prelude::*;
+
+// https://docs.python.org/3/howto/free-threading-python.html#identifying-free-threaded-python
+const GIL_DISABLED_PROBE: &str =
+    "import sysconfig; print(sysconfig.get_config_var('Py_GIL_DISABLED') or 0)";
```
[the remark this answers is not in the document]
not-matthias: No, we should expect to have a working environment. If we don't have that, the benches also won't work.

